PR #1, opened 2020-12-06
Fix go report card to use correct import path

Jacalz:
The link used the wrong import path. This should fix it 👍


README.md
```diff
@@ -2,7 +2,7 @@
   <a href="https://pkg.go.dev/fyne.io/fyne-x?tab=doc" title="Go API Reference" rel="nofollow"><img src="https://img.shields.io/badge/go-documentation-blue.svg?style=flat" alt="Go API Reference"></a>
   <a href='http://gophers.slack.com/messages/fyne'><img src='https://img.shields.io/badge/join-us%20on%20slack-gray.svg?longCache=true&logo=slack&colorB=blue' alt='Join us on Slack' /></a>
   <br />
-  <a href="https://goreportcard.com/report/fyne.io/fyne-x"><img src="https://goreportcard.com/badge/fyne.io/fyne-x" alt="Code Status" /></a>
```
andydotxyz, line 5: Thanks for noticing - I think you need to fix both occurrences

andydotxyz (changes requested):

README.md
Jacalz: Good catch. Doing this on my phone and didn’t notice that there were two links. Fixed now :)

andydotxyz (approved):

PR #4, opened 2020-12-20
Initial import of my GIF widget

andydotxyz:
A chance for apps to have animated components with no custom code.


AlbinoGeek (commented):
Just some code questions, I may be wrong however, so take with a grain of salt of course.

It is not my intention to second-guess you, haha.

widget/gif.go
```diff
@@ -0,0 +1,122 @@
...
+func (g *AnimatedGif) Start() {
+	buffer := image.NewNRGBA(g.dst.Image.Bounds())
+	draw.Draw(buffer, g.dst.Image.Bounds(), g.src.Image[0], image.Point{}, draw.Over)
+	g.dst.Image = buffer
+	g.dst.Refresh()
+
+	go func() {
+		g.stopping = false
+
+		for !g.stopping {
```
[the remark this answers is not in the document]
stuartmscott: Agreed.

@andydotxyz I've also made a gif renderer and found GIF's LoopCount spec quite unintuitive, posting the code here in the hope it is helpful;
```go


func GIFDecoder(source io.ReadCloser) (fyne.CanvasObject, error) {
	defer source.Close()
	gif, err := gif.DecodeAll(source)
	if err != nil {
		return nil, err
	}
	size := gif.Image[0].Bounds().Size()
	overpaintImage := image.NewRGBA(image.Rect(0, 0, size.X, size.Y))
	output := &canvas.Image{
		Image:    overpaintImage,
		FillMode: canvas.ImageFillOriginal,
	}
	go func() {
		draw.Draw(overpaintImage, overpaintImage.Bounds(), gif.Image[0], image.ZP, draw.Src)
		loop := func() {
			for c, srcImg := range gif.Image {
				draw.Draw(overpaintImage, overpaintImage.Bounds(), srcImg, image.ZP, draw.Over)
				canvas.Refresh(output)

				time.Sleep(time.Millisecond * time.Duration(gif.Delay[c]) * 10)
			}
		}
		switch gif.LoopCount {
		case 0:
			// Loop forever
			for {
				loop()
			}
		case -1:
			// Loop once
			loop()
		default:
			// Loop LoopCount+1 times
			for i := 0; i <= gif.LoopCount; i++ {
				loop()
			}
		}
	}()
	return output, nil
}
```

widget/gif.go
```diff
@@ -0,0 +1,122 @@
...
+func (g *gifRenderer) Destroy() {
+	g.gif.Stop()
+}
+
+func (g *gifRenderer) Layout(size fyne.Size) {
+	g.gif.dst.Resize(size)
+}
+
+func (g *gifRenderer) MinSize() fyne.Size {
+	return fyne.NewSize(1, 1)
```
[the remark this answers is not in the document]
andydotxyz: It should not derive from image bounds due to scaling issues.
SetMinSize is an interesting point. I can't honestly decide because other widgets do not do this, but this is very close to the Image type in usage...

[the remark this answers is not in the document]
stuartmscott: I tend to agree there needs to be some way to prop it up as most layouts will shrink to minsize. Ideally it could be based on the image bounds, but I'm not familiar with the scaling issues so that may not be an option. `canvas.Image` extends base so it provides `SetMinSize`, however this is more of a widget so perhaps it could follow `widget.Icon` and use `theme.InlineIconSize()` as the minimum?

widget/gif.go
```diff
@@ -0,0 +1,136 @@
...
+			for c, srcImg := range g.src.Image {
+				if g.stopping {
+					break
+				}
+				draw.Draw(buffer, g.dst.Image.Bounds(), srcImg, image.Point{}, draw.Over)
+				g.dst.Refresh()
+
+				time.Sleep(time.Millisecond * time.Duration(g.src.Delay[c]) * 10)
+			}
+			g.remaining--
```
stuartmscott, line 98: This will eventually underflow, so it would not quite loop forever, but loop for 2*MAX_INT times

widget/gif.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil
+}
+
+// CreateRenderer loads the widget renderer for this widget. This is an internal requirement for Fyne.
+func (g *AnimatedGif) CreateRenderer() fyne.WidgetRenderer {
+	return &gifRenderer{gif: g}
+}
+
+// Start begins the animation. The speed of the transition is controlled by the loaded gif file.
+func (g *AnimatedGif) Start() {
```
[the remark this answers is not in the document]
stuartmscott: The running flag should prevent this, though maybe it should checked and set inside a mutex

widget/gif.go
```diff
@@ -0,0 +1,136 @@
...
+	read, err := storage.Reader(u)
+	if err != nil {
+		return err
+	}
+	pix, err := gif.DecodeAll(read)
+	if err != nil {
+		return err
+	}
+	g.src = pix
+	g.dst.Image = pix.Image[0]
```
stuartmscott, line 56: Should `g.dst` be refreshed after setting the image?

README.md
```diff
@@ -23,4 +23,13 @@ Community contributed layouts.
 
 Community contributed widgets.
 
+### Animated Gif
+
+A widget that will run animated gifs.
+
+```go
+gif, err := NewAnimatedGif(storage.NewFileURI("./testdata/minions.gif"))
```
stuartmscott, line 31: What license is this / do we have permission from Universal Pictures to use this (or whoever owns the rights)? Might be better to use the rotating Earth from https://en.wikipedia.org/wiki/Gif which is under Creative Commons

stuartmscott (changes requested):

widget/gif.go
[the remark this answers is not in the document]
andydotxyz: >  I'm not familiar with the scaling issues

images are measured in pixels, but in vector land we don't know how many pixels per DP until runtime - so if you prop open based on image size your output will be very different sizes depending on output device.

I think I prefer SetMinSize over using IconSize (as GIF doesn't make a good icon ;) )

README.md
andydotxyz: I downloaded it from a royalty free site. rotating earth is probably better, thanks.

widget/gif.go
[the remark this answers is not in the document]
stuartmscott: :+1: SetMinSize sounds good to me

widget/gif.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Start begins the animation. The speed of the transition is controlled by the loaded gif file.
+func (g *AnimatedGif) Start() {
+	if g.isRunning() {
+		return
+	}
+	g.runLock.Lock()
+	defer g.runLock.Unlock()
+	g.running = true
```
stuartmscott, line 89: This doesn't need to be deferred
```suggestion
	g.running = true
	g.runLock.Unlock()
```

README.md
stuartmscott: 👍

This line needs to be updated;
```suggestion
gif, err := NewAnimatedGif(storage.NewFileURI("./testdata/gif/earth.gif"))
```

stuartmscott (changes requested):

README.md
andydotxyz: Thanks, looks like the search missed that occurrence!

stuartmscott (approved):

PR #6, opened 2020-12-28
Spinner Widget

AlbinoGeek:
# Spinner

image

<details><summary>Example</summary>
<p>

This Example demonstrates a few uses for the Spinner widget\.

```go
package main

import (
	"fyne.io/fyne"
	"fyne.io/fyne/app"
	"fyne.io/fyne/container"
	"fyne.io/fyne/layout"
	"fyne.io/fyne/widget"
)

func main() {
	var (
		a = app.NewWithID("x.fyne.demo.spinner")
		w = a.NewWindow("Spinner Demo")

		basic     = xwidget.NewSpinner(0, 1)
		integer   = xwidget.NewIntSpinner(0, 1)
		negative  = xwidget.NewSpinner(-5, 0.637)
		positive  = xwidget.NewSpinner(5, 0.637)
		precision = xwidget.NewSpinner(0, 0.28671058)
	)

	negative.SetMax(0)
	positive.SetMin(0)

	// shows the value as accurately as possible (default is to round to SpinnerDefaultPrecision)
	precision.SetPrecision(-1)

	w.SetContent(container.NewVBox(
		widget.NewForm(
			widget.NewFormItem("Basic", basic),
			widget.NewFormItem("Integer", integer),
			widget.NewFormItem("Max", negative),
			widget.NewFormItem("Min", positive),
			widget.NewFormItem("Precision", precision),
		),
		fyne.NewContainerWithLayout(
			layout.NewGridLayout(2),
			widget.NewButton("Disable", func() {
				basic.Disable()
				integer.Disable()
				negative.Disable()
				positive.Disable()
				precision.Disable()
			}),
			widget.NewButton("Enable", func() {
				basic.Enable()
				integer.Enable()
				negative.Enable()
				positive.Enable()
				precision.Enable()
			}),
		),
	))

	w.CenterOnScreen()
	w.ShowAndRun()
}
```

</p>
</details>

## Index

- [Variables](<#variables>)
- [type Spinner](<#type-spinner>)
  - [func NewIntSpinner(value, step int) *Spinner](<#func-newintspinner>)
  - [func NewSpinner(value, step float64) *Spinner](<#func-newspinner>)

## Variables

SpinnerDefaultPrecision controls the default number of digits shown after the decimal point in the Spinner's entry\. The special precision \-1 uses the smallest number of digits necessary such that the value is represented to the best of a float64's ability\.

### See Also

strconv\.ParseFloat

```go
var SpinnerDefaultPrecision = 12
```

## type Spinner

Spinner widget has a floating point value shown in a widget\.Entry\, with widget\.Button s to the right which increment or decrement the spinner value by a set step value\. Also provides an integer version\, a set minimum and/or maximum\, and visual precision for floationg point values\.

Note: Since this component uses float64 it should not be considered 100% mathematically accurate\. For more information\, please see: https://0.30000000000000004.com/\. If you need this\, then you would have to re\-implement this component using a decimal math library such as: github\.com/ericlagergren/decimal\. A conscious choice not to use this library was made due to the significant performance hit involved in using software over hardware floating point math\.

```go
type Spinner struct {
    fyne.Container
    // contains filtered or unexported fields
}
```

### func NewIntSpinner

```go
func NewIntSpinner(value, step int) *Spinner
```

NewIntSpinner creates a spinner with an integer initial value and step increment\.

### func NewSpinner

```go
func NewSpinner(value, step float64) *Spinner
```


AlbinoGeek:
# Question

Should the `go.mod` contain the indirect dependencies in this repository or not?

Reasons for not:

- It's a library, and specifying indirects will lead to their versions being locked.
- It is the application (e.g: library user's) responsibility to opt into updates.

Asking because I realize this compiles with a `go.mod` containing only:

```go
module fyne.io/x/fyne

go 1.15

require (
	fyne.io/fyne v1.4.2
)
```

However, my IDE is configured to update transitive dependencies on save, which caused the contents:

```go
module fyne.io/x/fyne

go 1.15

require (
	fyne.io/fyne v1.4.2
	golang.org/x/net v0.0.0-20201224014010-6772e930b67b // indirect
	golang.org/x/sys v0.0.0-20201223074533-0d417f636930 // indirect
	golang.org/x/text v0.3.4 // indirect
	gopkg.in/yaml.v2 v2.4.0 // indirect
)
```

andydotxyz:
Yes, it's probably a good idea to remove the indirects... That probably applies to the Fyne project as well, it is not something we had considered before

andydotxyz:
> var SpinnerDefaultPrecision = 12

I think it would be cleaner if this matched the `Step` functionality of slider, then we have consistency - that would be, for example, `0.0001` instead - which looks more intuitive and is not limited to "multiples of 10".

andydotxyz (changes requested):
One of the reasons this was not previously accepted was because the up/down button should not take up the space of two normal buttons. We would prefer it be in the same space as, for example, the password enable/disable or the icon used to show validation failure.

AlbinoGeek:
> > var SpinnerDefaultPrecision = 12
> 
> I think it would be cleaner if this matched the `Step` functionality of slider, then we have consistency - that would be, for example, `0.0001` instead - which looks more intuitive and is not limited to "multiples of 10".

With the `SpinnerDefaultPrecision` set to `-1`, in theory, the value shown in the field would always be a multiple of your `Step`.

e.g: `0.0001` -> `0.0002` -> `0.0003`

HOWEVER; because floating point maths and representation of floating point values (without including an entire `decimal` library) is... "inaccurate" -- I introduced this variable to let the developer control the "visual rounding" of the Spinner widget. You can also use the `SetPrecision` method on any Spinner to override this package-level default.

This changes the number of significant digits shown in the representation of the float within the Spinner's Entry.

https://golang.org/pkg/strconv/#FormatFloat

This is a one-to-one mapping of the `prec` variable in the call to `FormatFloat` used to show the value.

Please feel free to tell me if I have missed something here, or mis-understood your comment in regards to `Step` on the slider.

---

> One of the reasons this was not previously accepted was because the up/down button should not take up the space of two normal buttons. We would prefer it be in the same space as, for example, the password enable/disable or the icon used to show validation failure.

Unfortunately, because the `Button`'s minimize size is determined by the Theme's `Padding` and `InlineIconSize` I cannot implement the more visually-appealing vertically aligned buttons without outright overruling these two Theme settings, and copy-pasting the entirety of the `widget.Button` code into a new file to changes its' hard-coded minimum size.

If there is another way to do this, I am all ears.

andydotxyz:
> I introduced this variable to let the developer control the "visual rounding" of the Spinner widget.

Visual rounding is better performed by a format string surely? It is also more flexible.

> If there is another way to do this, I am all ears.

Yes there is, override the MinSize of an extended Button widget.

AlbinoGeek:
> Visual rounding is better performed by a format string surely? It is also more flexible.

The idea behind exposing that (and `SetPrecision`) was that you'd get full control over it. That being said, I made a mistake in setting this value to `12` -- I should have reversed that (where the default is `-1` instead.) -- please see the screenshot below on the changes.

> Yes there is, override the MinSize of an extended Button widget.

This produces visually unappealing buttons (due to the large inline icon.) -- I think I have found a happy medium by using the wrong icons (`MenuUp`/`MenuDown` v.s. `MoveUp`/`MoveDown`)

### Updated Screenshot

image

AlbinoGeek:
> One of the reasons this was not previously accepted was because the up/down button should not take up the space of two normal buttons. We would prefer it be in the same space as, for example, the password enable/disable or the icon used to show validation failure.

> ### Updated Screenshot
> 
> image

Would you prefer I move the buttons to be rendered as the `Entry.ActionItem` ?

andydotxyz:
> Would you prefer I move the buttons to be rendered as the `Entry.ActionItem` ?

How it's actually implemented should not matter much I don't think.
If the above version had the padding removed from between the buttons it might fit (if you compare it to a normal entry it looks like that one might be taller?).

Jacalz (commented):
Hi @AlbinoGeek. Are you still interested in continuing your work on this? I think that this would be a great addition to the fyne-x code base.

As for the implementation, I agree with the previous comments about having the buttons inside the entry (as part of the ActionItem). I also think the new tutorial about making an entry only allow numerical inputs might be of use for this, see https://developer.fyne.io/tutorial/numerical-entry.

piratecarrot:
This is a slight update for fyne v2 and makes use of generics and data binding
It is by no means well written, just slapped together. I don't have time to really put together code to contribute right now hence a quick comment and a gist.

https://gist.github.com/piratecarrot/3be6e4e5abbf08641078642c187744dc

lordofscripts:
> This is a slight update for fyne v2 and makes use of generics and data binding It is by no means well written, just slapped together. I don't have time to really put together code to contribute right now hence a quick comment and a gist.
> 
> https://gist.github.com/piratecarrot/3be6e4e5abbf08641078642c187744dc

Where is the original Spinner code that has the Up/Down spinner buttons on top of each other rather than side-by-side as in your code? I noticed the Spinner is no longer available and I am desperately looking for one.

andydotxyz:
> I noticed the Spinner is no longer available and I am desperately looking for one.

It's on my todo list, but that is a very long list. If anyone were able to sponsor it I'm sure we could deliver as part of v2.5.0

lordofscripts:
I found some code of `Spinner[T]` and had to patch it to get the up/down buttons stacked. Also had to do some quick mods because I needed a letter (rune type) (A-Z).

My remaining problem is that I also need it to present as Read Only which is not the same as Disabled. The Disabled mode makes the value unreadable to the casual user. Anyway, still trying, hoping to see the official spinner some day.

andydotxyz:
Disabled is not supposed to be unreadable. Sounds like a theme issue not a read-only vs disabled issue.

PR #7, opened 2021-02-06
add hex/7seg widget

charlesdaniels:
This PR adds a simple skeuomorphic "7 segment" or "hex display" widget. Screenshots are included in the README.md for the widget. It didn't seem like we had an established way to demo these widgets, so I created a demo program under `cmd/`.


widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,214 @@
...
+	seg1.StrokeWidth = float32(hexSegmentWidth / 2)
+	seg2.StrokeWidth = float32(hexSegmentWidth / 2)
+	seg3.StrokeWidth = float32(hexSegmentWidth / 2)
+	seg4.StrokeWidth = float32(hexSegmentWidth / 2)
+	seg5.StrokeWidth = float32(hexSegmentWidth / 2)
+	seg6.StrokeWidth = float32(hexSegmentWidth / 2)
+
+	pos := image.Pt(0, 0)
+
+	// pt0Center := image.Pt(pos.X+int(hexWidth/2.0+hexOffset), pos.Y+int((hexHeight-2.0*hexSegmentVLength)/2))
```
andydotxyz, line 117: Looks like old code

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,214 @@
...
+
+	pt6Center := image.Pt(pos.X+int(hexWidth/2.0), int(float32(pt0Center.Y)+hexSegmentVLength))
+	pt65 := image.Pt(int(float32(pt6Center.X)-(hexSegmentHLength/2)), pt6Center.Y)
+	pt61 := image.Pt(int(float32(pt6Center.X)+(hexSegmentHLength/2)), pt6Center.Y)
+
+	pt3Center := image.Pt(pos.X+int(hexWidth/2.0-hexOffset), int(float32(pt0Center.Y)+2*hexSegmentVLength))
+	pt34 := image.Pt(int(float32(pt3Center.X)-(hexSegmentHLength/2)), pt3Center.Y)
+	pt32 := image.Pt(int(float32(pt3Center.X)+(hexSegmentHLength/2)), pt3Center.Y)
+
+	// segment 0
```
andydotxyz, line 130: I think either this comment is not needed, or the `seg0` should be `segment0` if you worry it's not clear

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,214 @@
...
+	h.Refresh()
+}
+
+// Set updates the hex widget to show a specific number between 0 and 15, which
+// will be rendered in hexadecimal in 0...f. If the number is greater than 15,
+// it will be modulo-ed by 16.
+func (h *HexWidget) Set(val uint) {
+	val = val % 16
+
+	if val == 0 {
```
andydotxyz, line 181: Go prefers `switch` over lots of if/else

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,214 @@
...
+// refresh so the changes are visible to the user.
+func (h *HexWidget) UpdateSegments(segments uint8) {
+	h.segments = segments
+	h.Refresh()
+}
+
+// Set updates the hex widget to show a specific number between 0 and 15, which
+// will be rendered in hexadecimal in 0...f. If the number is greater than 15,
+// it will be modulo-ed by 16.
+func (h *HexWidget) Set(val uint) {
```
andydotxyz, line 178: Random thought, would it be nicer to use `rune` OR would you consider consts for the `a`, rather than developer passing `10`?

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,214 @@
...
+		segments: 0xff,
+	}
+
+	h.ExtendBaseWidget(h)
+	return h
+}
+
+// UpdateSegments changes the state of the segments and causes the widget to
+// refresh so the changes are visible to the user.
+func (h *HexWidget) UpdateSegments(segments uint8) {
```
andydotxyz, line 170: Should this be public?

Another random thought, would it be clearer if this took a `...uint`, then did the bit management inside this method? (maybe less of an issue if this is private)

andydotxyz (changes requested):
This is a cool widget.
Some random thoughts inline, though you'd need to address the build failures as well.

widget/hexwidget/hexwidget.go
charlesdaniels: df17679c1ae6d6f2f3ad755f0ab930a11340f923

widget/hexwidget/hexwidget.go
charlesdaniels: df17679c1ae6d6f2f3ad755f0ab930a11340f923

widget/hexwidget/hexwidget.go
charlesdaniels: I don't think so. It can't really handle arbitrary runes. Also, Go allows for hexadecimal values. You can give run `h.Set(0xa)` and it will do the right thing. This way makes it easy to slice apart a bigger number with masks and set each corresponding segment individually. Using rune would make it require string processing, and consts would mean the caller would have to look stuff up.

widget/hexwidget/hexwidget.go
charlesdaniels: df17679c1ae6d6f2f3ad755f0ab930a11340f923

widget/hexwidget/hexwidget.go
charlesdaniels: 4f2bad7055186ccbed2afbdedab52af2e729a95c

I actually decided to make this a lookup table. There is no reason to re-compute these values at runtime, and indexing into a slice like this should be O(1) - no need for any comparisons at runtime at all.

charlesdaniels:
I am working on ironing out some of the build failures, which I will hopefully finish mopping up tonight. I also addressed all the code review comments, AFAIK.

Aside from this, I realized there was no good reason not to expose the variables to change the height/color/slant, so I exposed those in the public API, and also added some sliders/color pickers in the demo.

andydotxyz:
I think the slant calculations may not be working quite right, as it goes out of the bounds
Screenshot 2021-03-01 at 11 13 41

widget/hexwidget/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Screenshots
+
+![](hexwidget_00abcdef.png)
+
+![](hexwidget_12345678.png)
+
+## Future Work
+
+* Support scaling the size of the hex widget.
+* Support changing the hex widget color.
```
andydotxyz, line 16: It looks like this is done

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+	)
+}
+
+func (h *hexRenderer) Layout(size fyne.Size) {
+}
+
+func (h *hexRenderer) ApplyTheme() {
+}
+
+func (h *hexRenderer) BackgroundColor() color.Color {
```
andydotxyz, line 64: This is not needed any more

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+
+	h.segmentObjects[0].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[1].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[2].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[3].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[4].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[5].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+	h.segmentObjects[6].(*canvas.Line).StrokeWidth = float32(hexSegmentWidth / 2)
+
+	pos := image.Pt(0, 0)
```
andydotxyz, line 81: Using fyne.Position would be better as you would not lose precision

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+
+// SetSlant changes the amount of "slant" i nthe hex widgets. The topmost
+// segment is offset by slant many virtual pixels to the right. A value of 0
+// means no slant at all.
+func (h *HexWidget) SetSlant(s float32) {
+	h.hexOffset = s
+	h.Refresh()
+}
+
+func setLineEndpoints(l *canvas.Line, pt1, pt2 image.Point) {
```
andydotxyz, line 177: I'd recommend `fyne.Position` so that it is more precisely placed

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+	(1<<3 | (1 << 4)),
+	(1 << 3),
+	(1<<0 | (1 << 1)),
+	(1<<0 | (1 << 1) | (1 << 2) | (1 << 5)),
+	(1<<0 | (1 << 5)),
+	(1<<1 | (1 << 2)),
+	(1<<1 | (1 << 2) | (1 << 3)),
+}
+
+// size in pixels of the hex widget
```
andydotxyz, line 36: These are not pixel sizes, the Fyne coordinate system is device independent

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+//     -----
+//    |     |
+//  4 |     | 2
+//    |  3  |
+//     -----
+type HexWidget struct {
+	widget.BaseWidget
+	segments uint8
+
+	// size in "pixels" of the hex widget
```
andydotxyz, line 134: Not pixels, sorry.
Also, why not just use `fyne.Size` for a single field?

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+
+// SetSize changes the size of the hex widget.
+func (h *HexWidget) SetSize(s fyne.Size) {
+	h.hexHeight = s.Height
+	h.hexWidth = s.Width
+	h.Refresh()
+}
+
+// SetSlant changes the amount of "slant" i nthe hex widgets. The topmost
+// segment is offset by slant many virtual pixels to the right. A value of 0
```
andydotxyz, line 170: Not sure that "virtual pixels" means much, sounds like pixels again.

widget/hexwidget/hexwidget.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// SetSize changes the size of the hex widget.
+func (h *HexWidget) SetSize(s fyne.Size) {
+	h.hexHeight = s.Height
+	h.hexWidth = s.Width
+	h.Refresh()
+}
+
+// SetSlant changes the amount of "slant" i nthe hex widgets. The topmost
```
andydotxyz, line 169:
```suggestion
// SetSlant changes the amount of "slant" in the hex widgets. The topmost
```

widget/hexwidget/hexwidget.go
andydotxyz: Also, for clarity, is it worth stating (if this is true?) that setting slant = height will give 45degree slope?

andydotxyz (changes requested):
Looks cool, but some docs and/or coordinates could be improved

andydotxyz:
One other random observation... The widget starts with no segments set, but this is a state it can never return to.
I'd say that either it should either start with "0" visible, or have a way to go back to "off".

widget/hexwidget/hexwidget.go
charlesdaniels: Done in 541512d. 

This code was ported from a different toolkit, hence why there are some non-idiomatic things.

widget/hexwidget/hexwidget.go
charlesdaniels: 541512d

widget/hexwidget/hexwidget.go
charlesdaniels: 1cb3743

widget/hexwidget/hexwidget.go
charlesdaniels: 1cb3743

widget/hexwidget/hexwidget.go
charlesdaniels: 2f597c1

widget/hexwidget/README.md
charlesdaniels: bcf4e7c

widget/hexwidget/hexwidget.go
charlesdaniels: bcf4e7c

widget/hexwidget/hexwidget.go
charlesdaniels: AFAIK, this is correct.

bcf4e7c

widget/hexwidget/hexwidget.go
charlesdaniels: I have decided that this should be public, because it addresses your [other comment](https://github.com/fyne-io/fyne-x/pull/7#issuecomment-787878110) and allows the user finer-grained control if they want it.

I'm not sold on using a list of bool/uint, because then storing 1 hex state requires using a slice, which adds a lot more overhead. Additionally, users on the happy path can use the `Set()` function.

charlesdaniels:
I believe this is once again ready for review at your leisure.

andydotxyz (approved):
Such a cool widget, thanks :)

PR #8, opened 2021-02-16
simple widget interface, base implementation incl. simple renderer

tehsphinx:
My first draft of the SimpleWidget implementation. I will create a few comments on the code where I have open questions and probably still have to adjust / extend things.

Any improvement suggestions are very welcome.

---

Basic idea: Be able to easily create new widgets without having to write a lot of boiler code.

With SimpleWidget (only) the following 3 things are necessary in a widget implementing it:
- widget struct must be based on `SimpleWidgetBase`
- `ExtendBaseWidget` must be called on widget creation
- `Build` function must be implemented (overwritten)

No renderer needs to be implemented, as it is part of SimpleWidgetBase implementation.

### DIFF
--- a/widget/example/simple_widget.go
+++ b/widget/example/simple_widget.go
@@ -1,0 +1,47 @@
+package example
+
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/x/fyne/widget"
+)
+
+// SampleWidget is a sample widget demonstrating the base structure of a
+// widget implementing SimpleWidget using SimpleWidgetBase.
+type SampleWidget struct {
+	widget.SimpleWidgetBase
+	propertyLock sync.RWMutex
+
+	Property int
+}
+
+// NewSampleWidget creates a new sample widget.
+func NewSampleWidget(property int) *SampleWidget {
+	return &SampleWidget{Property: property}
+}
+
+// Render renders the SampleWidget.
+func (s *SampleWidget) Render() (objects []fyne.CanvasObject, layout func(size fyne.Size)) {
+	// create objects needed for rendering and append them to the objects slice.
+	// (executed in Widget.CreateRenderer function)
+
+	return objects, func(size fyne.Size) {
+		// position, resize or otherwise update objects. (executed in WidgetRenderer.Layout)
+	}
+}
+
+// SetStateSafe sets or changes the state of a widget in a safe way. A Refresh
+// is triggered after the state changes have been applied.
+func (s *SampleWidget) SetStateSafe(setState func()) {
+	s.SimpleWidgetBase.SetStateSafe(&s.propertyLock, setState)
+}
+
+func (s *SampleWidget) getProperty() int {
+	s.propertyLock.RLock()
+	defer s.propertyLock.RUnlock()
+
+	return s.Property
+}
+
+// All other methods of a fyne.Widget can be overwritten as usual. E.g. MinSize, Resize, Refresh, etc.
--- a/widget/simple_renderer.go
+++ b/widget/simple_renderer.go
@@ -1,0 +1,57 @@
+package widget
+
+import (
+	"fyne.io/fyne/v2"
+)
+
+// newSimpleRenderer creates a new simpleRenderer.
+func newSimpleRenderer(widget SimpleWidget, objects []fyne.CanvasObject, layout func(size fyne.Size)) *simpleRenderer {
+	return &simpleRenderer{
+		widget:  widget,
+		objects: objects,
+		layout:  layout,
+	}
+}
+
+// simpleRenderer is a renderer providing the basic rendering functionality used
+// by SimpleWidgetBase.
+type simpleRenderer struct {
+	widget  fyne.Widget
+	objects []fyne.CanvasObject
+
+	layout func(size fyne.Size)
+}
+
+// MinSize returns the minimum size of the widget that is rendered by this renderer.
+func (s *simpleRenderer) MinSize() fyne.Size {
+	return fyne.Size{Width: 0, Height: 0}
+}
+
+// Refresh is a hook that is called if the widget has updated and needs to be redrawn.
+// This might trigger a Layout.
+func (s *simpleRenderer) Refresh() {
+	s.Layout(s.widget.Size())
+}
+
+// Layout is a hook that is called if the widget needs to be laid out.
+// This should not be overwritten.
+func (s *simpleRenderer) Layout(size fyne.Size) {
+	s.layout(size)
+}
+
+// Objects returns the objects that should be rendered.
+//
+// Implements: fyne.WidgetRenderer
+func (s *simpleRenderer) Objects() []fyne.CanvasObject {
+	return s.objects
+}
+
+// SetObjects updates the objects of the renderer.
+func (s *simpleRenderer) SetObjects(objects []fyne.CanvasObject) {
+	s.objects = objects
+}
+
+// Destroy does nothing in the base implementation.
+//
+// Implements: fyne.WidgetRenderer
+func (s *simpleRenderer) Destroy() {}
--- a/widget/simple_widget.go
+++ b/widget/simple_widget.go
@@ -1,0 +1,117 @@
+package widget
+
+import (
+	"sync"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/widget"
+)
+
+// SimpleWidget defines an interface for fyne widgets that are simple
+// to implement when based on SimpleWidgetBase.
+// For more info on how to implement, see the documentation of the
+// SimpleWidgetBase.
+type SimpleWidget interface {
+	fyne.Widget
+
+	Render() ([]fyne.CanvasObject, func(size fyne.Size))
+}
+
+// SimpleWidgetBase defines the base for a SimpleWidget implementation.
+// To create a new widget base it on SimpleWidgetBase using composition.
+// Create a `New` function initialising the widget and make sure to call
+// ExtendBaseWidget in the New function. Always use the `New` function to
+// create the widget or make sure `ExtendBaseWidget` is called elsewhere.
+//
+// Overwrite the `Render() (objects []fyne.CanvasObject, layout func(size fyne.Size))`
+// function. It returns the (base-)objects needed to render the widgets content,
+// as well as a function `layout` responsible for positioning and resizing the
+// different objects based on the incoming available space for the widget.
+// Try not to define new objects in the `layout` function as they would be
+// recreated every time the widget is refreshed.
+//
+// Other functions defined by the fyne.Widget interface can be overwritten
+// and will be used by the SimpleWidgetBase if overwritten.
+//
+// See ./example/simple_wigdet.go for a bootstraped widget implementation.
+type SimpleWidgetBase struct {
+	widget.BaseWidget
+
+	propertyLock sync.RWMutex
+	impl         SimpleWidget
+}
+
+// Render must be overwritten in a widget to create other widgets and
+// canvas objects the widget is composed of. It returns a slice of the
+// created objects and the layout function.
+// The layout fucntion should be used to position and size the objects (widgets
+// and canvas objects). New objects should be created in the Render function body
+// outside the returned layout function, so they are not re-created
+// every time the widget gets refreshed.
+func (s *SimpleWidgetBase) Render() (objects []fyne.CanvasObject, layout func(size fyne.Size)) {
+	return nil, func(fyne.Size) {}
+}
+
+// CreateRenderer implements the Widget interface. It creates a simpleRenderer
+// and returns it. No renderer needs to be implemented. If the simpleRenderer
+// doesn't do it, SimpleWidget is probably not suitable for the use case.
+// Usually this should not be overwritten or called manually.
+func (s *SimpleWidgetBase) CreateRenderer() fyne.WidgetRenderer {
+	wdgt := s.super()
+	objs, layout := wdgt.Render()
+
+	return newSimpleRenderer(wdgt, objs, layout)
+}
+
+// SetState sets or changes the state of a widget. A Refresh
+// is triggered after the state changes have been applied.
+func (s *SimpleWidgetBase) SetState(setState func()) {
+	setState()
+	s.super().Refresh()
+}
+
+// SetStateSafe sets or changes the state of a widget in a safe way. A Refresh
+// is triggered after the state changes have been applied.
+// The provided sync.Locker should be the same you use for read protection of the
+// widget properties.
+func (s *SimpleWidgetBase) SetStateSafe(m sync.Locker, setState func()) {
+	m.Lock()
+	setState()
+	m.Unlock()
+
+	s.super().Refresh()
+}
+
+// ExtendBaseWidget is used by an extending widget to make use of BaseWidget functionality.
+func (s *SimpleWidgetBase) ExtendBaseWidget(wid SimpleWidget) {
+	impl := s.getImpl()
+	if impl != nil {
+		return
+	}
+
+	s.BaseWidget.ExtendBaseWidget(wid)
+
+	s.propertyLock.Lock()
+	defer s.propertyLock.Unlock()
+	s.impl = wid
+}
+
+func (s *SimpleWidgetBase) super() SimpleWidget {
+	impl := s.getImpl()
+	if impl == nil {
+		var x interface{} = s
+		return x.(SimpleWidget)
+	}
+	return impl
+}
+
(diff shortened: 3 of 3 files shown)


widget/simple_renderer.go
```diff
@@ -0,0 +1,57 @@
...
+
+// SetObjects updates the objects of the renderer.
+func (s *simpleRenderer) SetObjects(objects []fyne.CanvasObject) {
+	s.objects = objects
+}
+
+// Destroy does nothing in the base implementation.
+//
+// Implements: fyne.WidgetRenderer
+func (s *simpleRenderer) Destroy() {}
```
tehsphinx, line 57: I'm not yet sure when this functions needs to destroy something. We could provide the possibility to overwrite the `Destroy` function in the widget as well, if that is a common (or even seldom) use case.

widget/simple_renderer.go
```diff
@@ -0,0 +1,57 @@
...
+	widget  fyne.Widget
+	objects []fyne.CanvasObject
+
+	layout func(size fyne.Size)
+}
+
+// MinSize returns the minimum size of the widget that is rendered by this renderer.
+func (s *simpleRenderer) MinSize() fyne.Size {
+	return fyne.Size{Width: 0, Height: 0}
+}
```
tehsphinx, line 28: Do we need to be able to overwrite the MinSize function in the renderer, if it is already overwritten in the widget? If yes, I can make the renderer return the same as the MinSize function of the widget if overwritten.

widget/simple_renderer.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Refresh is a hook that is called if the widget has updated and needs to be redrawn.
+// This might trigger a Layout.
+func (s *simpleRenderer) Refresh() {
+	s.Layout(s.widget.Size())
+}
```
tehsphinx, line 34: Are there use cases where this is not enough? Should we be able to overwrite the `Refresh` function? If those cases are rare, I'd say for those cases one can fall back to implementing the `fyne.Widget` the usual way.

widget/simple_widget.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Build must be overwritten in a widget. It returns a slice of child objects
+// and a layout function.
+// The layout function should be used to position and size the objects (widgets
+// and canvas objects). New objects should be created in the Build function body,
+// so they are not re-created every time the widget gets refreshed.
+func (s *SimpleWidgetBase) Build() (objects []fyne.CanvasObject, layout func(size fyne.Size)) {
+	return nil, func(fyne.Size) {}
+}
```
tehsphinx, line 51: I renamed `Renderer` to `Build`. Is there less potential for confusion?

widget/simple_widget.go
```diff
@@ -0,0 +1,115 @@
...
+// is triggered after the state changes have been applied.
+// The provided sync.Locker should be the same you use for read protection of the
+// widget properties.
+func (s *SimpleWidgetBase) SetStateSafe(m sync.Locker, setState func()) {
+	m.Lock()
+	setState()
+	m.Unlock()
+
+	s.super().Refresh()
+}
```
tehsphinx, line 81: helper functions to update state without having to call `Refresh` manually.

widget/example/simple_widget.go
```diff
@@ -0,0 +1,49 @@
...
+	return objects, func(size fyne.Size) {
+		// position, resize or otherwise update objects. (executed in WidgetRenderer.Layout)
+	}
+}
+
+// SetStateSafe sets or changes the state of a widget in a safe way. A Refresh
+// is triggered after the state changes have been applied.
+func (s *SampleWidget) SetStateSafe(setState func()) {
+	s.SimpleWidgetBase.SetStateSafe(&s.propertyLock, setState)
+}
```
tehsphinx, line 40: Convenience function to avoid having to pass in the mutex every time.

widget/example/simple_widget.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (s *SampleWidget) getProperty() int {
+	s.propertyLock.RLock()
+	defer s.propertyLock.RUnlock()
+
+	return s.Property
+}
```
tehsphinx, line 47: sample function to show mutex protection on property read

tehsphinx:
Could potentially solve https://github.com/fyne-io/fyne/issues/709

widget/simple_renderer.go
AlbinoGeek: Destroy often goes unused in widgets, even custom ones -- however, now that data binding is in the framework (as of 2.0.0) it is possible that a widget may need to do cleanup.

widget/simple_renderer.go
AlbinoGeek: `Refresh` doesn't have to call `Layout` in most cases, this could reduce performance if it's the default behaviour. (However, I definitely don't mind the idea of a `RefreshAfterLayout` sort of option, if you ever decide to go down the options route.)

AlbinoGeek:
Looks like an interesting idea, will be curious to see where this goes.

widget/simple_renderer.go
tehsphinx: ok, I'll provide a way to be able to implement/overwrite it.

widget/simple_renderer.go
tehsphinx: So, if I understand correctly, `Refresh` is called for e.g. data updates and `Layout` needs to be called for resizing/repositioning? Or what exactly is the definition/distinction?

Since the goal is to simplify widget implementation, it kinda feels complicated to me to have to understand the difference and implement 3 different functions (CreateRenderer, Layout and Refresh). I'd like to keep it with 2 (CreateRenderer and Layout). Or do you see another option?
I'm not sure I got the `RefreshAfterLayout` option. Don't see how that would help keeping it simple.

tehsphinx: Does Layout trigger a rebuild all the way down the widget tree even if there is no position/size change?

tehsphinx: Refresh can now be overwritten and will default to calling Layout.

widget/simple_renderer.go
tehsphinx: Destroy can now be overwritten if needed.

tehsphinx:
From my side this suggestion is finished.

I would add tests after we've agreed on moving forward with this.

widget/simple_renderer.go
tehsphinx: @andydotxyz?

tehsphinx: MinSize can now be overwritten in widget and will be passed down to the renderer.

andydotxyz: Most widgets have a min size defined by their renderring - and so the default widget `MinSize` implementation simply delegates to the renderer.
Remember that we want complete separation between widget and render output (in the API design).

widget/simple_renderer.go
andydotxyz: > So, if I understand correctly, `Refresh` is called for e.g. data updates and `Layout` needs to be called for resizing/repositioning? Or what exactly is the definition/distinction?

That is the destinction. Layout is very expensive as it requires checking the size of all child objects.
Refresh is faster because it changes properties and pushes the rest of the work to renderer/GPU.

widget/simple_renderer.go
andydotxyz: > I'm not yet sure when this functions needs to destroy something.

When widgets are not visible their renderers could be removed, in which case they will be asked to clean up with `Destroy`.

widget/simple_widget.go
andydotxyz: This is indeed better - I wonder if it is coincience that the return type here is basically that of a containers :)

widget/simple_widget.go
andydotxyz: The former of the two is implicityly `SetStateUnsafe` - should we be exposing unsafe methods?

andydotxyz:
It seems like there are two separate things here, a simple widget wrapper, and a safe state management API.
The latter is more likely to be moved upstream than the former.

tehsphinx:
I can remove the state management API from this PR if that pulls the focus away from the simple widget wrapper, which this PR should be about mainly. What do you think? 🤔

andydotxyz:
Thinking more about this today it seems that this `SimpleWidget` is actually not simplifying widgets, but widget renderers. In fact because you’ve pushed the `MinSize` code back into widget this could be seen as more complicated than before.

I wonder if instead this belongs as a `SimpleWidgetRenderer` which is where its value seems to lie. Then any widget could return this from `CreateRenderer` and we don’t need to document different types of base widget?...

tehsphinx:
It does not simplify widget or renderer. It simplifies _**widget creation**_.

## Before, ...
Before, someone that wants to create a new widget (which I want to do like multiple times a day -- at least I do when writing flutter code, React code or Vue code) had to:
- create a struct implementing Widget interface
- base it on BaseWidget
- create a CreateRenderer function returning another struct
- create a renderer struct implementing WidgetRenderer
- implement 5 (!) functions on the renderer

As a newcomer there are the additional hurdles of:
- understanding that some of the renderer code actually goes into the widgets CreateRenderer method (creation of objects).
- understanding why there is Layout and Refresh and what code should go where.
- understanding that Destroy can be left empty in most cases
- understanding what to return from MinSize as it needs some return value
- understanding what to return from Objects.

## With SimpleWidget
- one struct needs to be created based on SimpleWidgetBase
- one function needs to be implemented with a clear description on what to do
- optionally more functions can be overwritten as needed, but don't get in the way for a newcomer as they default to a working solution for most cases

tehsphinx:
I think this is a matter of "outside perspective" vs "inside perspective".

As someone very new to fyne I knew within the hour that I wanted a simpler way to create widgets. I was sure there was already something and when I didn't find it, I went ahead and started to create my own.

Creating widgets just needed too much boiler plate code and too much looking at existing widgets to even understand what to do. The documentation on widget creation helped but there are too many missing pieces.

Also I think it will help adoption a lot if there is a way to do things that is similar to other modern GUI frameworks out there. That's why I'm always mentioning flutter, react and vue, as there are many people out there that know how it's done there and it's always somewhat similar.

In case this is not clear: I don't want to replace anything. I also don't mean the existing code to change in any way. This is helper code building ON TOP of the existing framework making it a lot easier for people to create new widgets. And again, not widgets that replace the existing ones, but custom combinations of existing widgets that build up to higher level widgets for the very specific use cases the users have.

andydotxyz:
I understand your and appreciate the summary. However (if we omit render details) extending `BaseWidget` is almost exactly the same as this `SimpleWidget` - you make a new struct, extend a thing and implement a single required method to return the render details.

So this is what I mean by the indication that this is largely easing rendering - if `CreateRenderer` could return `NewSimpleWidgetRenderer(...)` we can avoid needing new widget types... Is that not true?

You are right that we should be inspired by other GUI frameworks but I don't think we should mimic their APIs. We can document how they differ if that will help - but our plan was to use a fresh start to find a better solution - we will not be well-served by trying to match other ways unless we think it matches our design.

tehsphinx:
From a users (developer who uses fyne) point of view: Why is there Widget and WidgetRenderer? If I want to create a new widget, why do I need to create 2 structs?

(I think), I know why, from the frameworks point of view. The widget contains the definition with the properties and the renderer is created by the framework as needed and "renders" the widget to the canvas.  It can be thrown away and recreated.

But from a users point of view, there is no reason to split the two. So, I would not be happy with a SimpleWidgetRenderer, because the widget part would still be mostly boilerplate.

andydotxyz:
> Why is there Widget and WidgetRenderer?

Because Widget API should be focused on behaviour and logical state - not rendering.
The rendering of a widget is separate from it's API which is why they are separate types. Any visual information that is cached should be in the renderer instead of the widget. Renderers have a different lifecycle than widgets as they are only needed when displayed.

andydotxyz:
> because the widget part would still be mostly boilerplate.

I don't agree that the following is too much boilerplate:

```go
type myWidget struct {
	widget.BaseWidget
}

func (m *myWidget) CreateRenderer() fyne.WidgetRenderer {
	return NewSimpleWidgetRenderer(...)
}
```

tehsphinx:
Ok, so going with your reasoning: Where would you create the child widgets and canvas elements that this widget is made off?

Because as is, this doesn't hold:
> Because Widget API should be focused on behaviour and logical state - not rendering.

All the widgets I have seen so far put rendering details in the `CreateRenderer` function which is part of the widget.

andydotxyz:
> All the widgets I have seen so far put rendering details in the CreateRenderer function which is part of the widget.

Yes we got lazy. Really it should be in a render constructor function. This is a shorthand we adopted that avoids adding a new method.
If they were in two separate files the `CreateRenderer` method would exist with the renderer rather than the widget for better grouping of code.

tehsphinx:
The split between widget and renderer only creates problems and extra hurdles for the implementer.

I tried that approach already before. I have now added a sample implementation. Because in the renderer you'd want access to the state and the properties of the widget, a reference to the widget is needed. It can't be abstracted (at least not without generics) as you'd not want to have to type cast the widget. This leaves the CreateRenderer function (although pretty much predefined) to the developer to be implemented.

While I like the idea of keeping the state and properties separate from the renderer, it just doesn't turn out to be practical enough for the user. This could be revisited in 1 year when generics are out, but since you seem to care about some backwards compatibility, it is more like 3 years (static check indicates compatibility with go1.12 which was released 2 years ago).

AlbinoGeek:
> While I like the idea of keeping the state and properties separate from the renderer, it just doesn't turn out to be practical enough for the user. This could be revisited in 1 year when generics are out, but since you seem to care about some backwards compatibility, it is more like 3 years (static check indicates compatibility with go1.12 which was released 2 years ago).

AFAIK the only reason they support `Go1.12` is because some version of `Mac OS X` which still exists for some reason **only ships this version**, and it's not immediately obvious to the user of that OS version as to how they would get a newer version.

E.g.: `fyne` seems to wish to support any LTS OS's package manager's version of Golang.

AlbinoGeek:
> I don't agree that the following is too much boilerplate:
> 
> ```go
> type myWidget struct {
> 	widget.BaseWidget
> }
> 
> func (m *myWidget) CreateRenderer() fyne.WidgetRenderer {
> 	return NewSimpleWidgetRenderer(...)
> }
> ```

If only a real-world usage for widget wrapping were every this simple though (as mentioned above with the reference passing.)

I didn't even know for the longest time that I wasn't supposed to keep state in my Widget, simply because there was **no obvious breakage** for doing so, e.g.: The framework didn't stop me.

If there was a method that a developer could call to crash/rebuild the Driver for example... then we could actually see what happens when our state gets messed up, and start to care about that separation as much as the core developers do.

(This would be equiv. to `go test -race` in the sense that it's testing an unusual condition of the application, and could be a build tag that caused the application to perform much slower, etc, as long as it would help developers understand the need for proper state separation.)

Don't get me wrong (please don't take offense by my wording), when I say "we don't care", I mean "it doesn't seem like we should, because we don't understand the internal reasoning for it, and it's not made evidently apparent through the API or documentation."

E.g.: There is very little penalty for "doing it wrong" atm. E.g: If I keep 100% of my state and logic in the `Widget`, and only create a dummy renderer, it may be "incorrect form", but it's not like it doesn't work. In fact, it works just fine, and I haven't noticed a performance or stability difference between a widget which manages its state within the renderer vs the widget itself.

If this difference was more pronounced, we would better understand the need for separation.

andydotxyz:
> The split between widget and renderer only creates problems and extra hurdles for the implementer.

I disagree. The split is essential for separation of these concerns, cleaner testing, and better maintainability over time.
It is also key to avoiding accidentally having render state cached in widget objects which can never be cleared from memory if there is a reference, for example, to hidden elements.

> Because in the renderer you'd want access to the state and the properties of the widget, a reference to the widget is needed.

Yes, a renderer can reference the widget for state, that is intended. However a widget should not reference a renderer.

andydotxyz:
> This is what makes Renderers currently feel like a "dangling implementation detail" of Widgets.

Yes I can see that. As we progress towards Aberlour release we will see further disconnection of renderers, to enable better thread safety in rendering. We will also start to see more renderers being garbage collected at runtime.

> If there was a method that a developer could call to crash/rebuild the Driver for example..

This is an interesting thought, something that we could indeed consider adding.

andydotxyz:
I am struggling with this ticket a lot, apologies. I have outline our design and why it is the way that it is.
I don't think it likely that we will merge in widgets or utilities that do not match our design (API or UI).
We can have a separate discussion about the shortcomings of our design and how to improve them, but a PR is probably not the best place to do that. We have a proposals process that allows us to spend time weighing the benefits and drawbacks of larger changes. https://github.com/fyne-io/fyne/wiki/Contributing%3A-Proposals

PR #11, opened 2021-04-18
Propose a completion entry

metal3d:
This is a completion entry that proposes a popup menu below the entry and that can be refreshed while the user types.

Navigation keys are done to make it possible to the user to use Up and Down arrow to select the chosen value.

Tests are complicated, I talked to Andrew about the problem. When the popup menu appears, the focus is "lost" and I cannot send the typed keys to the correct component in tests. I made several tries with several implementations without success. The component works but is not fully testable.


widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+	items []string
+	pause bool
+}
+
+// NewCompletionEntry creates a new CompletionEntry which creates a popup menu that responds to keystrokes to navigate through the items without losing the editing ability of the text input..
+func NewCompletionEntry(options []string) *CompletionEntry {
+	c := &CompletionEntry{}
+	// c.ExtendBaseWidget(c)
+	c.Entry = widget.NewEntry()
+	c.items = options
```
Jacalz, line 33: You can add this to the original struct like so:
```go
c := &CompletionEntry{items: options}
```

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+	cmenu *widget.PopUpMenu
+	items []string
+	pause bool
+}
+
+// NewCompletionEntry creates a new CompletionEntry which creates a popup menu that responds to keystrokes to navigate through the items without losing the editing ability of the text input..
+func NewCompletionEntry(options []string) *CompletionEntry {
+	c := &CompletionEntry{}
+	// c.ExtendBaseWidget(c)
+	c.Entry = widget.NewEntry()
```
Jacalz, line 32:
```suggestion
	c.ExtendBaseWidget(c)
```
This should work, I think.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+	c.pause = false
+	return c
+}
+
+// ShowCompletion displays the completion menu
+func (c *CompletionEntry) ShowCompletion() {
+	if c.pause {
+		return
+	}
+	items := make([]*fyne.MenuItem, 0)
```
Jacalz, line 43: Why make the slice to zero length? I think it would be better to make it to the length of items. Then you can workaround having to append each time. Should be a bit faster.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)
+	c.cmenu.ShowAtPosition(fyne.Position{X: pos.X, Y: pos.Y + c.Size().Height})
+}
+
+// HideCompletion hides the completion menu.
+func (c *CompletionEntry) HideCompletion() {
+	if c.cmenu == nil {
+		return
+	}
+	c.cmenu.Hide()
```
Jacalz, line 69:
```suggestion
	if c.cmenu != nil {
		c.cmenu.Hide()
	}
```

Jacalz (changes requested):
Just some comments on code improvement. I did unfortunately not have the time to look more closely into the usage and such.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+// Example:
+//
+//    entry := NewCompletionEntry()
+//    entry.OnChanged = func(s string) {
+//        if len(s) < 3 {
+//            entry.HideCompletion()
+//        }
+//        entry.SetItems([]string{"foo", "bar", "baz"}
+//        entry.ShowCompletion()
+//    }
```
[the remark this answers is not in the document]
metal3d: I'll do it, no problem

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+type CompletionEntry struct {
+	*widget.Entry
+	menu  *fyne.Menu
+	cmenu *widget.PopUpMenu
```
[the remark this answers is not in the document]
metal3d: oh... I squashed the bad commit, leaving the old one where I didn't changed the variable name yet... Sorry for that

widget/entry-with-completion.go
```diff
@@ -0,0 +1,141 @@
...
+	pause bool
+}
+
+// NewCompletionEntry creates a new CompletionEntry which creates a popup menu that responds to keystrokes to navigate through the items without losing the editing ability of the text input..
+func NewCompletionEntry(options []string) *CompletionEntry {
+	c := &CompletionEntry{}
+	// c.ExtendBaseWidget(c)
+	c.Entry = widget.NewEntry()
+	c.items = options
+	c.pause = false
```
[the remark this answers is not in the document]
metal3d: right !

widget/entry-with-completion.go
metal3d: right too...

widget/entry-with-completion.go
metal3d: That's a problem, if I do this, there are poblems when I try to get the canvas for object, and to pass events to the entry. I will make some tries in a couple of minutes

widget/entry-with-completion.go
metal3d: I mimic what it's done in the EntrySelect from Fyne. I agree that I can use the size of items to prepare the capacity.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,127 @@
...
+	c.pause = true
+	c.Entry.SetText(s)
+	c.Entry.Refresh()
+	c.pause = false
+}
+
+// FocusGained make the Entry to be focused.
+//
+// Implements: fyne.Focusable
+func (c *CompletionEntry) FocusGained() {
```
fpabl0, line 74: When you embed a type, you don't need to redefine its functions, unless you need to change its behavior.

fpabl0, line 74: When you embed a type, you don't need to redefine its functions, unless you need to change its behavior.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,127 @@
...
+	c.Entry.FocusGained()
+}
+
+// FocusLose is only here to respect interface.
+//
+// Implements: fyne.Focusable
+func (c *CompletionEntry) FocusLose() {}
```
fpabl0, line 81: Same here

widget/entry-with-completion.go
```diff
@@ -0,0 +1,127 @@
...
+
+// TypedRune sends the rune to the Entry, whatever if the popup menu for completion is displayed or not.
+//
+// Implements: fyne.Focusable
+func (c *CompletionEntry) TypedRune(r rune) {
```
fpabl0, line 86: same

widget/entry-with-completion.go
```diff
@@ -0,0 +1,127 @@
...
+	c.Entry.TypedRune(r)
+}
+
+// TypedKey sends the key events to the Entry, whatever if the popup menu for completion is displayed or not.
+//
+// Implements: fyne.Focusable
+func (c *CompletionEntry) TypedKey(event *fyne.KeyEvent) {
```
fpabl0, line 93: same

widget/entry-with-completion.go
metal3d: Resolved, don't know why that failed before

widget/entry-with-completion.go
```diff
@@ -0,0 +1,127 @@
...
+	c.Entry.TypedKey(event)
+}
+
+// Bind event and rune key to the Entry if needed, and bind othters navigation key + validation with Return.
+// This bound keys are set when the "popupmenu" gains the focus, the canvas continues to send rune and keys
+// to the Entry.
+func (c *CompletionEntry) bindKey(cnv fyne.Canvas) {
+	position := -1
+	cnv.SetOnTypedRune(func(r rune) {
```
fpabl0, line 102: I am unsure about this. Canvas callbacks could be edited at application level by the developer. So by doing this, there would be a chance that the developer change this in its application (then the completion entry would not work), or viceversa, the completion entry would override the developer setting.

fpabl0 (changes requested):
Some little suggestions on the code :)

widget/entry-with-completion.go
metal3d: If I remove this function, the entry lose the focus

widget/entry-with-completion.go
metal3d: For this one, yes, I forget to remove it since I now use ExtendBaseWidget, I'll do it right now

widget/entry-with-completion.go
metal3d: no, really I need to leave it because the Entry lose the focus while the popup menu is opened

widget/entry-with-completion.go
metal3d: and same for this one, I cannot remove it

widget/entry-with-completion.go
metal3d: The entry loses the focus while the popup menu is displayed, there is no other way to manage navigation for popup menu **and** continue to edit the entry without having key bound at canvas levelr :( I spent hours to find solutions, speaking with Andrew and trying maybe 10 implementations. This is the only one that works correctly :)

widget/entry-with-completion.go
Jacalz: What I meant was that there is no point in using `make()` to create an empty slice. It is better to just use `var items []*fyne.MenuItem` in that case. SelectEntry actually does prepare the capacity as of https://github.com/fyne-io/fyne/pull/2169. Thanks for updating it to use that.

widget/entry-with-completion.go
metal3d: Finally, this one is not useful

widget/entry-with-completion.go
metal3d: OK, this one is not useful because I have the focus while the popup menu is not displayed

Jacalz (changes requested):
The image looks like the drop down is too small. The popUp width should match that of Select and SelectEntry (be as long as the entry).

widget/entry-with-completion.go
metal3d: OK... While I now use ExtendBaseWidget it becaime useless. Sorry...

fpabl0: I think that happened because you didn't ExtendBaseWidget previously. But now that you correctly extends it this should work

metal3d:
I can resize the popup menu to the entry, you're right

widget/entry-with-completion.go
fpabl0: Yes, I know the reason behind this, but unfortunately this would lead to that kind of problems :(

metal3d: In my opinion, it's a shame to prevent yourself from having this type of widget simply because the basic library suffers from some design issues.
We would be depriving ourselves of offering some very useful tools, really very useful, because Fyne doesn't offer some routines.
Here the widget works, potentially we could add a note that this widget cannot be overloaded at the canvas level.
But this widget is one of the first ones that I complained about that it doesn't exist... And I admit that since version 1 of Fyne, I don't expect a major API change that allows us to have completion on an Entry...

metal3d: (sorry, I used the word "shame" because I was helped by a translator... and didn't seen that word. I would say "a pity", I fixed it)

fpabl0: I am with you, I really like this new widget :). But maybe we should find a solution for this first. For example, canvas could add listeners to that events instead of set (replace) them.

metal3d: It's also possible to mark this as "experimental" until something changes in Fyne to allow me to not use the Canvas to bind keys... But really, this widget is a big miss in Fyne, and I though that fyne-x could propose it ;)

metal3d: "But maybe we should find a solution for this first. For example, canvas could add listeners to that events instead of set (replace) them."
I don't see what you mean. Is this not what does the "TypedKey" and "TypedRune" ?

fpabl0: > I don't see what you mean. Is this not what does the "TypedKey" and "TypedRune" ?

I mean this `canvas.SetOnTypedRune` could add a functionality instead of replace it.

By the way, I think this could be possible to do without using canvas bindings, maybe a custom popup menu that listen to typed runes or keys?

metal3d: That's what I tried before. The problem is not the popupmenu but that it is in a secondary canvas (overlay). If I do not do this, I cannot place the menu because I don't have the container where to place the "menu".
SetOnTypedRune failed for me, I lose the cursor in the Entry (and I don't know why).
If you've got a begining of what I could change (because I really don't see other way right now), that could be nice. I will take a look.

fpabl0: Ok I have done some tests, and it is possible to do without canvas bindings :). However, currently is not possible to do with the current version of Fyne. Because this requires to be able to focus manually objects in an overlay which I think it is not possible to do outside the fyne repository (internal implementation). But that should be added soon I think :)

fpabl0: >  The problem is not the popupmenu but that it is in a secondary canvas (overlay). 

Absolutely. There should be a function to focus overlay content objects. Sorry, I wrote the above comment, without realized that you had answered.

andydotxyz: > Because this requires to be able to focus manually objects in an overlay

this should be possible. The code changed on `develop` but it should have worked before as well.

I agree that a component such as this needs to use focused based key events instead of canvas hooks. If it is not working/possible then we have a bug that should be fixed

fpabl0: Introducing that function, would allow us to do this without canvas bindings. However this time, you should implement manually the menu (that should implement `fyne.Focusable`) with a widget.PopUp (not a PopUpMenu).

https://user-images.githubusercontent.com/12239342/115155041-9bad2200-a043-11eb-86b1-5c2f5fd48d4e.mp4

metal3d: But anyway I will need to put it in the overlay, are we OK ?

fpabl0: @andydotxyz 
> If it is not working/possible then we have a bug that should be fixed

Yes, this should be fixed, basically the problem is that each overlay has its own focus manager.

> But anyway I will need to put it in the overlay, are we OK ?

Yes, absolutely. You should create the menu manually and then pass it to `widget.NewPopUp(menu, canvas)`.

metal3d: OK, I'll do it in a couple of hours, thanks for your time and idea. I hope to propose this to the community, Fyne and Fyne-X are really cool stuffs :)

fpabl0: > thanks for your time and idea

On the contrary thanks to you, this is a great addition :)

fpabl0: @metal3d
Have just realized that it is indeed possible to do it currently, as @andydotxyz said.

When you already build the menu manually and insert it to the widget.PopUp, you can focus the menu immediately after it is shown by doing:
```go
c.popup = widget.NewPopUp(menu, holder)
c.popup.Resize(fyne.NewSize(c.Entry.Size().Width, c.popup.MinSize().Height))

pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)
c.popup.ShowAtPosition(fyne.Position{X: pos.X, Y: pos.Y + c.Size().Height})
holder.Focus(menu) // 👈 added
```
Note that `holder.Focus` needs to be called just after `popup.ShowAtPosition(...)` because the overlay only exist when it is shown.

metal3d: This is not the wanted behavior because the cursor disappears from the entry

fpabl0: @metal3d Did you try it? The cursor will not disappear because overlays have their own focus scope, so by doing `holder.Focus(menu)`, the entry will remain focused.

metal3d: I tried.
There is a problem with `widget.Menu` (overriden in a popup) that won't take the size of the popup and cannot be "updated", so I'm now using a List (with data bind). 
I implemented the Focusable interface, I print a log in `FocusGained`.

With a simple example: no problem (there is a problem with the List height but let's check the basics)

```go
func (n *navigableList) FocusGained() {
	log.Println("Focused !")
}
```

In test:
```go
	entry := createEntry()
	win := test.NewWindow(entry)
	win.Resize(fyne.NewSize(500, 300))
	defer win.Close()

	entry.SetText("init") // here the "Focused !" log appears from FocusGained
	entry.ShowCompletion() // => displays <nil>
	elem := test.Canvas().Overlays().Top()
	log.Println("Focused => ", elem)
```

```
=== RUN   TestCompletionEntry_ShowMenu
2021/04/19 10:32:43 Focused !
2021/04/19 10:32:43 Focused =>  <nil>
```

I tried to get `win.Canvas().Focus()` instead of Overlay => same, it's "<nil>".

Whatever I try, it works on app, but the test won't give me the focused object.

I also tried:

```go
    test.Canvas().Focus(entry)
```
Same result...

metal3d: This is my implementation:

```go
package widget

import (
	"log"

	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/data/binding"
	"fyne.io/fyne/v2/widget"
)

// CompletionEntry is an Entry with options displayed in a PopUpMenu.
type CompletionEntry struct {
	widget.Entry
	popupMenu     *widget.PopUp
	navigableList *navigableList
	items         []string
	pause         bool
}

// NewCompletionEntry creates a new CompletionEntry which creates a popup menu that responds to keystrokes to navigate through the items without losing the editing ability of the text input..
func NewCompletionEntry(options []string) *CompletionEntry {
	c := &CompletionEntry{items: options}
	c.ExtendBaseWidget(c)
	// c.Entry = widget.NewEntry()
	return c
}

// ShowCompletion displays the completion menu
func (c *CompletionEntry) ShowCompletion() {
	if c.pause {
		return
	}

	if c.navigableList == nil {
		c.navigableList = newNavigableMenu(c.items, &c.Entry, c.setTextFromMenu)
	}
	holder := fyne.CurrentApp().Driver().CanvasForObject(c)
	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)

	if c.popupMenu == nil {
		c.popupMenu = widget.NewPopUp(c.navigableList, holder)
	}
	c.popupMenu.ShowAtPosition(fyne.Position{X: pos.X, Y: pos.Y + c.Size().Height})
	c.popupMenu.Resize(fyne.NewSize(c.Entry.Size().Width, c.Size().Height*float32(len(c.items))))
	holder.Focus(c.navigableList)
}

// HideCompletion hides the completion menu.
func (c *CompletionEntry) HideCompletion() {
	if c.popupMenu != nil {
		c.popupMenu.Hide()
	}
}

// SetOptions set the completion list with itemList.
func (c *CompletionEntry) SetOptions(itemList []string) {
	c.items = itemList
	if c.navigableList != nil {
		c.navigableList.SetOptions(itemList)
	}
}

// Prevent the menu to open when the user validate value from the menu.
func (c *CompletionEntry) setTextFromMenu(s string) {
	c.pause = true
	c.Entry.SetText(s)
	c.Entry.Refresh()
	c.pause = false
	c.popupMenu.Hide()
}

type navigableList struct {
	*widget.List
	data            binding.StringList
	entry           *widget.Entry
	selected        int
	setTextFromMenu func(string)
	navigating      bool
}

func newNavigableMenu(items []string, entry *widget.Entry, setTextFromMenu func(string)) *navigableList {
	n := &navigableList{entry: entry, selected: -1, setTextFromMenu: setTextFromMenu}
	n.data = binding.BindStringList(&items)

	n.List = widget.NewListWithData(n.data, func() fyne.CanvasObject {
		return widget.NewLabel("")
	}, func(i binding.DataItem, o fyne.CanvasObject) {
		o.(*widget.Label).Bind(i.(binding.String))
	})

	n.OnSelected = func(id int) {
		if !n.navigating {
			item, _ := n.data.GetItem(id)
			val, _ := item.(binding.String).Get()
			n.entry.SetText(val)
		}
		n.navigating = false
	}

	return n
}

func (n *navigableList) SetOptions(items []string) {
	n.Unselect(n.selected)
	n.data.Set(items)
	n.selected = -1
}

func (n *navigableList) TypedKey(event *fyne.KeyEvent) {
	n.entry.TypedKey(event)
	switch event.Name {
	case fyne.KeyDown:
		if n.selected < n.data.Length()-1 {
			n.selected++
		}
		n.navigating = true
		n.List.Select(n.selected)

	case fyne.KeyUp:
		if n.selected > 0 {
			n.selected--
		} else {
			n.Unselect(n.selected)
			n.selected = -1
		}
		n.navigating = true
		n.List.Select(n.selected)
	case fyne.KeyReturn:
		n.navigating = false
		n.OnSelected(n.selected)
	}
}
func (n *navigableList) TypedRune(r rune) {
	n.entry.TypedRune(r)
}
func (n *navigableList) FocusGained() {
	log.Println("Focused !")
}
func (n *navigableList) FocusLost() {}

```

metal3d: And dont take care of names, I will change it later before to refresh the PR

andydotxyz: > Yes, this should be fixed, basically the problem is that each overlay has its own focus manager.

They do indeed, but calling `canvas.Focus` will find the correct manager.

andydotxyz: > This is not the wanted behavior because the cursor disappears from the entry

This is going to be the main challenge, because the "top most" item must have focus (so you can navigate the list) but you want it to look like the entry has focus. This may be possible by extending the entry to encourage it to look focused despite not being. However it may not be possible - it may require a little brainstorming with the team as well because I think the right balance is not obvious.

metal3d: > They do indeed, but calling `canvas.Focus` will find the correct manager.

This is what I did but the focus is not set. Is this a problem with Fyne ?

metal3d: Ho... It seems that win.Canvas().Focused() returned now the entry... So... Maybe...

andydotxyz: > This is what I did but the focus is not set. Is this a problem with Fyne ?

I don't know without looking at some code in detail, but I suspect the code should work or things like the Form dialog would not work. Maybe you did not add the overlay before calling focus like @fpabl0 suggested?

metal3d:
Don't be afraid, as soon as we are OK I will rebase the branch. Here I'm using a list because of the widget.Menu cannot be refreshed while the user is typing and because its size is not reset. List is OK, but:
- height is set to an arbitrary value (items count * entry size)
- there are 2 shadows, one for the popup and the other for List - that's ungly
- keyboard doesn't trigger scroll when an item is selected...

Should I make a complete list from scratch ?

widget/entry-with-completion.go
metal3d: No worry, it's now OK. I don't know what happens earlier but the tests are now OK. Excepting that I don't like the "list" usage (see my last commits)

Jacalz:
> Don't be afraid, as soon as we are OK I will rebase the branch. Here I'm using a list because of the widget.Menu cannot be refreshed while the user is typing and because its size is not reset. List is OK, but:
> 
> - height is set to an arbitrary value (items count * entry size)
> 
> - there are 2 shadows, one for the popup and the other for List - that's ungly
> 
> - keyboard doesn't trigger scroll when an item is selected...
> 
> 
> 
> Should I make a complete list from scratch ?

I am already working on something similar. https://github.com/fyne-io/fyne/pull/2181 It's just in early stages as of now though.

metal3d:
Arrrrrggg ... So I did it for nothing :(

Jacalz:
> Arrrrrggg ... So I did it for nothing :(

No, I did no completion. I mean that I just use the List widget for the select drop down.

metal3d:
Ah :)

For the heigh, maybe I can propose to set the number of elements to display in the constrictor (as a choice), I found a calculation that is not so bad.

The other problem (that is very bad) is the scroll that won't work on keyboard navigation.

This is mainly because the List proposes a scrollbar... I wonder if a "NoScrollList" widget would make the work a bit simpler.

OR ! If there is a way to go back to the "menu" that can be refreshed and correctly resized...

(I must admit that this widget is very important for me to finalize a demo app that I want to share in a lightning talk for the company where I work... I can avoid this of course, but that's a pitty 😉  )

widget/entry-with-completion.go
fpabl0: > This is going to be the main challenge, because the "top most" item must have focus (so you can navigate the list) but you want it to look like the entry has focus.

@andydotxyz it should not be hard, as they do not share the focus manager, so the entry will be focused in content focus manager, and the navigateList will be focused in the top most overlay focus manager.

README.md
```diff
@@ -49,3 +49,31 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
 <p align="center" markdown="1" style="max-width: 100%">
   <img src="img/widget-filetree.png" width="1024" height="880" alt="FileTree Widget" style="max-width: 100%" />
 </p>
+
+### CompletionEntry
+
+An extension of widget.Entry for displaying a popup menu for completion. Keyboard keys can be used to navigate (up and down) and select the option to set to the Entry (Enter key) - the mouse can be used too.
+
+```go
+entry := widget.CompletionEntry([]string{})
```
fpabl0, line 58: I think here should be:
```go
entry := widget.NewCompletionEntry([]string{})
```

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+func (c *CompletionEntry) setTextFromMenu(s string) {
+	c.pause = true
+	c.Entry.SetText(s)
+	c.Entry.Refresh()
+	c.pause = false
+	c.popupMenu.Hide()
+}
+
+type navigableList struct {
+	*widget.List
```
fpabl0, line 72: When you are extending a widget, you should embed the type as a value and not as pointer.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+	setTextFromMenu func(string)
+	hide            func()
+	navigating      bool
+}
+
+func newNavigableList(items []string, entry *widget.Entry, setTextFromMenu func(string), hide func()) *navigableList {
+	n := &navigableList{entry: entry, selected: -1, setTextFromMenu: setTextFromMenu, hide: hide}
+	n.data = binding.BindStringList(&items)
+
+	n.List = widget.NewListWithData(n.data, func() fyne.CanvasObject {
```
fpabl0, line 85: Taking the above suggestions, this should be rewritten to:
```go
n.List = widget.List{
	Length: func() int {
		return n.data.Length()
	},
	CreateItem: func() fyne.CanvasObject {
		return widget.NewLabel("")
	},
	UpdateItem: func(i widget.ListItemID, o fyne.CanvasObject) {
		item, err := n.data.GetItem(i)
		if err != nil {
			fyne.LogError(fmt.Sprintf("Error getting data item %d", i), err)
			return
		}
		o.(*widget.Label).Bind(item.(binding.String))
	},
}

n.ExtendBaseWidget(n) // 👈 this line is very important
```
With this change, the scroll should work when navigating with keys.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+		return widget.NewLabel("")
```
fpabl0, line 86: As you may notice, `widget.Label` introduces a lot of padding, so if we really want to make this looking as before, there should be a custom widget that uses a `canvas.Text`. But I think this is optional and it is subjective 😅

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (n *navigableList) TypedKey(event *fyne.KeyEvent) {
+	n.entry.TypedKey(event)
+	switch event.Name {
+	case fyne.KeyDown:
+		if n.selected < n.data.Length()-1 {
+			n.selected++
+		}
+		n.navigating = true
+		n.List.Select(n.selected)
```
fpabl0, line 117: `n.Select` should be enough as you are embedding the list :)

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+	n.data.Set(items)
+	n.selected = -1
+}
+
+func (n *navigableList) TypedKey(event *fyne.KeyEvent) {
+	n.entry.TypedKey(event)
+	switch event.Name {
+	case fyne.KeyDown:
+		if n.selected < n.data.Length()-1 {
+			n.selected++
```
fpabl0, line 114: Maybe should be better, if the n.selected resets to 0, when it goes until the end:
```go
if n.selected < n.data.Length()-1 {
	n.selected++
} else {
	n.selected = 0
}
```

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		n.navigating = true
+		n.List.Select(n.selected)
+
+	case fyne.KeyUp:
+		if n.selected > 0 {
+			n.selected--
```
fpabl0, line 121: Same suggestion here, maybe should be better if it select the last item when try to navigate up the first item:
```go
if n.selected > 0 {
	n.selected--
} else {
	n.selected = n.data.Length() - 1
}
```

widget/entry-with-completion.go
```diff
@@ -0,0 +1,140 @@
...
+		} else {
+			n.Unselect(n.selected)
+			n.selected = -1
+		}
+		n.navigating = true
+		n.List.Select(n.selected)
```
fpabl0, line 127: n.Select should be enough as you are embedding the list :)

fpabl0 (changes requested):
This is looking really good :). Just some suggestions.

README.md
metal3d: Ooops...

widget/entry-with-completion.go
metal3d: The problem is to calls the ListWithData construtor... If List is a value, and use *NewListWidtData, to setup the widget element, I've got a lot of side effects... I will try some things in a few minutes

widget/entry-with-completion.go
metal3d: Oh ok... thanks

widget/entry-with-completion.go
metal3d: I'm not sure. If the user scroll down and press Down, while the List won't autoscroll, it will lose the selection cursor. I would change this when the List from Fyne does the autoscroll, isn't it ?

widget/entry-with-completion.go
metal3d: and same remark 😉

widget/entry-with-completion.go
metal3d: right, I'll do it in a few minutes

widget/entry-with-completion.go
metal3d: Actually... You're right. The scroll problem would be resolved later but the common completion list behavior is to "rotate" in the list.

fpabl0: That's right. However I wonder if the completion list really need the scroll feature. For example, in chrome the input autocompletion does not scroll it always show a fixed length list of options. Also, scrolling a large list of options isn't a good user experience, is it? 🤔

metal3d:
I will rebase, there ate too many trashable commits.

fpabl0 (commented):
This comment could be removed 😅 I just wrote it to focus your attention on it:
```go
// 👈 this line is very important
```
It looks good, thanks :). I think @Jacalz may help you with the popup size (it is too large when you have a lot of options), as he is working on something similar.

metal3d:
LOL My god, Tired... I fix it :)

fpabl0 (dismissed):
LGTM :) thanks. Sorry for all requested changes. Just one more suggestion, next time please do not force-push commits as it is difficult to review (we have to see all the code again as everything appears as it were new)

metal3d:
A big thanks for your help @fpabl0 - hoping that the next one will be not so complicated to validate 😉

Jacalz:
I'll try to have a look at this when I get the chance.

README.md
```diff
@@ -49,3 +49,31 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
...
+    if len(s) < 3 {
+        // for example, you can avoid to display the completion
+        // if the entry has a value lenght < 3
+        entry.HideCompletion()
+        return
+    }
+
+    // here, you can get compltion from a dataset then
+    // set the list to the completion
+    entry.SetOptions([]string{"foo", "bar", "baz"})
```
Jacalz, line 69: I don’t know if it is just me, but it seems odd to create the entry with no items at all and then add them in the OnChanged function.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,156 @@
...
+	entry           *widget.Entry
+	selected        int
+	setTextFromMenu func(string)
+	hide            func()
+	navigating      bool
+}
+
+func newNavigableList(items []string, entry *widget.Entry, setTextFromMenu func(string), hide func()) *navigableList {
+	n := &navigableList{entry: entry, selected: -1, setTextFromMenu: setTextFromMenu, hide: hide}
+	n.data = binding.BindStringList(&items)
```
Jacalz, line 85: I think you might be better off not using bindings. There is a slight performance penalty when using it and I think the code could be a bit cleaner without it. You can see https://github.com/fyne-io/fyne/pull/2181 for some samples of how I did it. That applies to the list code below as well.

Jacalz (changes requested):
I did not have access to my development computer to run the code, but I had a quick read through and offered a few suggestions.

README.md
metal3d: That was to simplify the example, but if you want I can change the example.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,156 @@
...
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+// CompletionEntry is an Entry with options displayed in a PopUpMenu.
+type CompletionEntry struct {
+	widget.Entry
+	popupMenu     *widget.PopUp
+	navigableList *navigableList
+	items         []string
```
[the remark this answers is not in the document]
metal3d: If you want. But I don't think it's a good practice, it is possible that in the future we need to do some actions when items changed. A method is, IMHO better.
But, I will expose it. In a couple of minutes.

widget/entry-with-completion.go
metal3d: I use binding for a good reason 😄 

The main difference between SelectEntry and CompltionEntry is that I need to change the list content while the user type, without recreating it.

At this time, there is no "easy" way to reset the List items to change it.

Maybe the example I give in README is not representative, I have another example that I need:

- when user type, I do a "completion search" on a database (or API)
- I must remove the entire elements and rewrite them
- I should **not** close and open the list, I need to change the content

This capture shows the result:

https://user-images.githubusercontent.com/100080/115305424-9268a600-a166-11eb-9983-c8ad50110cc6.mp4

Or maybe you've got a nicer solution to clean the list and reset items?

README.md
metal3d: Finally, I don't understand the problem you point here. The completion is commonly dynamically changed while the user type text. So it's normal to have an empty completion and to change it on change.

Here, the completion is static, so that could be badly understood. But keep in mind an example of completing the word from an API or database while user type.

I can probably give more explanation in the documentation.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,156 @@
...
+				fyne.LogError(fmt.Sprintf("Error getting data item %d", i), err)
+				return
+			}
+			o.(*widget.Label).Bind(item.(binding.String))
+		},
+	}
+
+	n.ExtendBaseWidget(n)
+
+	n.OnSelected = func(id int) {
```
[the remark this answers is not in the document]
metal3d: To precise, "OnSelect" is part of `List`, so I only apply my function to the already declared method.

README.md
Jacalz: Alright. I might have misunderstood your meaning with this. I thought it just was supposed to auto complete from a given set of options passed to it.

widget/entry-with-completion.go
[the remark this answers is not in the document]
Jacalz: Yes it is part of the list, which I think means that you should be able to put it into the `widget.List{}` struct above. Your `n` variable is created before that.

README.md
metal3d: So yes, my example is bad :) I will provide another one that is more explicit.

widget/entry-with-completion.go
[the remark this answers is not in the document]
metal3d: Ah right, sorry, I made so many changes that I thought that everything was made in the `n` var declaration.

README.md
metal3d: I changed the example with a completion on Wikipedia API (and set the screenshot corresponding to  the result)

metal3d:
This is the actual example.

demo-completion

metal3d:
@Jacalz I will rebase to master as soon as you think it's OK. Thanks for your review 😄

Jacalz:
Thanks, I'll try to have a look at it later today on actual hardware. I don't think there is any need to rebase. We can easily squash and merge.

metal3d:
Rebase is easy while I'm working in a branch (git rebase -i master => I select what to sqash, commonly the entire commits excepting the first, then I validate).

Thanks again.

Jacalz:
> Rebase is easy while I'm working in a branch (git rebase -i master => I select what to sqash, commonly the entire commits excepting the first, then I validate).
> 
> 
> 
> Thanks again.

We mostly prefer that PRs are not rebased and then subsequently force pushed after the PR has been reviewed because it breakers review comments on GitHub. We'll squash and merge instead.

README.md
```diff
@@ -49,3 +49,56 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
...
+    }
+
+    // Make a search on wikipedia
+    resp, err := http.Get(
+        fmt.Sprintf(
+            "https://en.wikipedia.org/w/api.php?action=opensearch&search=%s", entry.Text,
+        ),
+    )
+    if err != nil {
+        entry.HideCompletion()
```
andydotxyz, line 75: If you return here like above then the remaining code can avoid being inside an else.
We generally prefer this so that the "main path" is top level in a method.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,154 @@
...
+		c.navigableList = newNavigableList(c.Options, &c.Entry, c.setTextFromMenu, c.HideCompletion)
+	}
+	holder := fyne.CurrentApp().Driver().CanvasForObject(c)
+	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)
+
+	if c.popupMenu == nil {
+		c.popupMenu = widget.NewPopUp(c.navigableList, holder)
+	}
+	max := fyne.Min(10, float32(len(c.Options)))
+	c.popupMenu.Resize(fyne.NewSize(c.Entry.Size().Width, c.popupMenu.MinSize().Height*max-3*theme.Padding()))
```
andydotxyz, line 41: Is this height calculation correct? In the screenshot it looks to be too long...

andydotxyz: The *3 seems a little random. And it isn't accounting for spacers, perhaps it needs `-(max-1)*theme.SeparatorThicknessSize()`

widget/entry-with-completion.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// SetOptions set the completion list with itemList and update the view.
+func (c *CompletionEntry) SetOptions(itemList []string) {
+	c.Options = itemList
+	c.Update()
+}
+
+// Update the list to refresh the options to display.
```
andydotxyz, line 59: Why add the new `Update()` when `Refresh()` is already in the API?

andydotxyz (changes requested):

andydotxyz:
This is looking really cool :) Just a few questions about API and the height of the drop-down.

README.md
```diff
@@ -49,3 +49,56 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
...
+    if len(s) < 3 {
+        entry.HideCompletion()
+        return
+    }
+
+    // Make a search on wikipedia
+    resp, err := http.Get(
+        fmt.Sprintf(
+            "https://en.wikipedia.org/w/api.php?action=opensearch&search=%s", entry.Text,
+        ),
```
Jacalz, line 72: Not really any need to use fmt here. Just append to the string, should be much faster and cleaner.

README.md
```diff
@@ -49,3 +49,56 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
...
+    )
+    if err != nil {
+        entry.HideCompletion()
+    } else {
+        // Get the list of possible completion
+        results := make([][]interface{}, 0)
+        dec := json.NewDecoder(resp.Body)
+        dec.Decode(&results)
```
Jacalz, line 80: We are not using `dec` anywhere else, I guess you could just do `json.NewDecoder(resp.Body).Decode(&results)` if you wanted to.

Jacalz (changes requested):
Thanks. This is now looking really close. Sorry for all the back and forward.
There seems to be issues regarding the size of the popup being bigger than the screen and not moving as it should when the entry is moved. Perhaps similar to https://github.com/fyne-io/fyne/pull/2183.
fyne-x-completion-entry

widget/entry-with-completion.go
metal3d: I renamed it after a few minutes because I was afraid that Refresh overrides something on the base widget... I will make the change and check.

widget/entry-with-completion.go
metal3d: I will try

README.md
metal3d: Ha, that was because I made some other tests and forget to remove the block.

widget/entry-with-completion.go
metal3d: this fails for me, the completion is too small in height

metal3d: c.popupMenu.MinSize().Height*max-1*theme.SeparatorThicknessSize() seems to be ok

widget/entry-with-completion.go
metal3d: I needed to Refresh entry so, otherwise the cursor disappears from the entry

metal3d:
@Jacalz yes, I remark this and I don't find any solution. I'm not sure that the problem comes from my widget...

metal3d:
> We mostly prefer that PRs are not rebased and then subsequently force pushed after the PR has been reviewed because it breakes review comments on GitHub. We'll squash and merge instead.

Ok no problem :)

metal3d:
> > @Jacalz yes, I remark this and I don't find any solution. I'm not sure that the problem comes from my widget...
> 
> I think the same solution in the PR that I linked to should work for you as well (when it comes to popup movement). You just have to overwrite the Move() method and move the popup as well.

I will try this, thanks a lot !

andydotxyz:
> I will try this, thanks a lot !

Can I hold off re-review until that change is on then?

metal3d:
I'm currently making the change @andydotxyz

metal3d:
@Jacalz It doesn't work for me. I tried to force the menu to a certain Y position (constant) to check. Le popup is well placed before the Move method is called and then it goes over the entry. Logging the `popupPos() `returns values gives me `{0, 37}` for example, but moving the menu doesn't change anything.

Note: moving the list makes a kind of "padding" on top, so it's placed below the entry, the only problem is the popup wich goes to 0 instead of the Y value.

```go
// Implements: fyne.Widget
func (e *CompletionEntry) Move(pos fyne.Position) {
	e.Entry.Move(pos)
	log.Println(pos, e.popUpPos())
	if e.popupMenu != nil {
		e.popupMenu.Move(e.popUpPos())
	}
}

func (c *CompletionEntry) popUpPos() fyne.Position {
	entryPos := fyne.CurrentApp().Driver().AbsolutePositionForObject(&c.Entry)
	return entryPos.Add(fyne.NewPos(0, c.Size().Height))
}

```

==> 

```
2021/04/22 14:23:00 {0 0} {0 37}
2021/04/22 14:23:00 {0 0} {0 37}
2021/04/22 14:23:00 {0 0} {0 37}
2021/04/22 14:23:00 {0 0} {0 37}
2021/04/22 14:23:00 {0 0} {0 37}
2021/04/22 14:23:00 {0 0} {0 37}
```

image
 
image

metal3d:
I finally found a way... I place the popup and resize it to cover the available space. So it is flexible now. The problem came from its size it seems.

Jacalz (changes requested):
I am sorry about all the back and forth on this. I really should have seen this earlier. Just two small things that I noticed. Apart from that, it is as good as approved from my part.

README.md
```diff
@@ -49,3 +49,52 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
...
+
+    // Make a search on wikipedia
+    resp, err := http.Get(
+        "https://en.wikipedia.org/w/api.php?action=opensearch&search=" + entry.Text,
+    )
+    if err != nil {
+        entry.HideCompletion()
+    }
+    // Get the list of possible completion
+    results := make([][]interface{}, 0)
```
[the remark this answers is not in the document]
Jacalz: Should be much faster and cleaner :slightly_smiling_face:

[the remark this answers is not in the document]
metal3d: And it's also more readable

Jacalz (approved):
Nice work. Thank you very much for this :+1:

metal3d:
No problem, your review was nice and I hope that this will help.


BTW: there are many of you advice that we can apply to the other source inside fyne-x 😄 😛

widget/entry-with-completion.go
```diff
@@ -0,0 +1,176 @@
...
+// calculate where the popup should appear
+func (c *CompletionEntry) popUpPos() fyne.Position {
+	entryPos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)
+	return entryPos.Add(fyne.NewPos(0, c.Size().Height))
+}
+
+// calculate the max size to make the popup to cover everything below the entry
+func (c *CompletionEntry) maxSize() fyne.Size {
+	cnv := fyne.CurrentApp().Driver().CanvasForObject(c)
+	return fyne.NewSize(c.Size().Width, cnv.Size().Height-c.Position().Y-c.Size().Height-theme.InputBorderSize()-theme.Padding())
```
andydotxyz, line 74: This should not draw over the whole canvas.
It should limit the height to the size of the list - otherwise it covers the entire area of the app below the entry (often with a lot of empty space).

andydotxyz (changes requested):

andydotxyz:
It looks good, but in testing I realised that it covers the whole of the app when this is used at the top of a window. 
Lots of empty space under the last item in the list.

https://user-images.githubusercontent.com/294436/116248446-cb56db00-a763-11eb-9a0e-ec4de47aaf16.mov

metal3d:
Absolutely @andydotxyz - I didn't found any solution to fix the bug that makes the list (popup) covering the entry when its size is higher than the window size.

Without changing the size (to an arbitrary size) of the canvas, it moves to 0,0 without any reason and cover the entry - see the problem reported by @Jacalz here https://github.com/fyne-io/fyne-x/pull/11#pullrequestreview-641403154

I decided to use the window size (minus the entry position) to ensure that it will not cover the entry. It's probably a bad choice, I admit...

andydotxyz:
> I decided to use the window size (minus the entry position) to ensure that it will not cover the entry. It's probably a bad choice, I admit...

I do not think that we are talking about the same thing. I was requesting that the popup maximum height be the size of it's content, so we do not cover the UI with an empty popup.

metal3d:
@andydotxyz actually we are talking about the same thing. I had too many problem with the popup covering the entry whatever the size I set. But I will try another time to fix this.

widget/entry-with-completion.go
metal3d: It's resolved in https://github.com/fyne-io/fyne-x/pull/11/commits/d82cdaef8120e90209180b7fede7d19e5ca14279 if I did things well ;)

widget/entry-with-completion.go
```diff
@@ -0,0 +1,191 @@
...
+func (c *CompletionEntry) popUpPos() fyne.Position {
+	entryPos := fyne.CurrentApp().Driver().AbsolutePositionForObject(c)
+	return entryPos.Add(fyne.NewPos(0, c.Size().Height))
+}
+
+// calculate the max size to make the popup to cover everything below the entry
+func (c *CompletionEntry) maxSize() fyne.Size {
+	cnv := fyne.CurrentApp().Driver().CanvasForObject(c)
+
+	if c.itemHeight == -1 {
```
Jacalz, line 76: This will brake usacases where users don't use the constructor, I think. Can't you just leave it at 0?

metal3d: yep

metal3d:
It's not OK for the max size. There were a few problems with the small completion list that made a useless scrollbar appearing.

andydotxyz:
> I had too many problem with the popup covering the entry whatever the size I set. But I will try another time to fix this.

The way that overlays work is that if you specify a position/size combination that would move it out of the window the driver will try to shuffle them on to screen.
If you want to avoid this consider checking what the available height is and asking for that, or less (hence why I was suggesting limit the max size).
Of course if you start adding custom code for moving overlays you may want to consider what if this completion is at the bottom of the screen... Should it position above the entry instead?

metal3d:
> > I had too many problem with the popup covering the entry whatever the size I set. But I will try another time to fix this.
> 
> The way that overlays work is that if you specify a position/size combination that would move it out of the window the driver will try to shuffle them on to screen.
> If you want to avoid this consider checking what the available height is and asking for that, or less (hence why I was suggesting limit the max size).

Yes, I did the calculation and it's now working. I had some troubles finding the item height but your advice to create a temporary item worked ;)

> Of course if you start adding custom code for moving overlays you may want to consider what if this completion is at the bottom of the screen... Should it position above the entry instead?

That's what I wonder too. At this time it's not the case, because I didn't decide the exact behavior - For example, what if the entry is positioned at 2/3 of the "screen" height... 

And what if the entry is in a scrolled view...

For now, it works if the entry has got enough space below the entry, either it is out of screen. I wanted to enhance the behavior later...

andydotxyz:
> Limiting to a certain number of items (so the height) is not common in other graphic libraries, but why not...

I do not understand what you mean here. Surely you are not suggesting that other toolkits will automatically stretch empty drop-down to the end of the screen?
I am curious because I don't think that anything I proposed is strange or radical - I just don't think we should waste space.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,191 @@
...
+	listheight := float32(len(c.Options))*(c.itemHeight+2*theme.Padding()+theme.SeparatorThicknessSize()) + 2*theme.Padding()
+
+	if cnv.Size().Height > listheight {
+		return fyne.NewSize(c.Size().Width, listheight)
+	}
+
+	return fyne.NewSize(
+		c.Size().Width,
+		cnv.Size().Height-c.Position().Y-c.Size().Height-theme.InputBorderSize()-theme.Padding())
+}
```
fpabl0, line 90:
```suggestion
      canvasSize = cnv.Size()
      entrySize = c.Size()
	if canvasSize.Height > listheight {
		return fyne.NewSize(entrySize.Width, listheight)
	}

	return fyne.NewSize(
		entrySize.Width,
		canvasSize.Height-c.Position().Y-entrySize.Height-theme.InputBorderSize()-theme.Padding())
}
```

fpabl0 (changes requested):
Just a little suggestion :)

metal3d:
> > Limiting to a certain number of items (so the height) is not common in other graphic libraries, but why not...
> 
> I do not understand what you mean here. Surely you are not suggesting that other toolkits will automatically stretch empty drop-down to the end of the screen?
> I am curious because I don't think that anything I proposed is strange or radical - I just don't think we should waste space.

There is no waste space in my latest commits - I fixed it.

I probably didn't understand what you mean here:
> If you want to avoid this consider checking what the available height is and asking for that, or less (hence why I was suggesting limit the max size).

Maybe because you said this after my fixes 😄

andydotxyz:
> I probably didn't understand what you mean here:
> 
> > If you want to avoid this consider checking what the available height is and asking for that, or less (hence why I was suggesting limit the max size).
> 
> Maybe because you said this after my fixes 😄

It is also possible that you read my message after fixing it.

andydotxyz:
There are a bunch of compile errors, no idea how this worked locally even.

fpabl0:
> There are a bunch of compile errors, no idea how this worked locally even.

It was my fault. I forgot to use ":=" in my suggestion. Sorry.

fpabl0 (changes requested):
Sorry, I made a mistake in my previous suggestion :(. Hope this time everything goes well.

widget/entry-with-completion.go
```diff
@@ -0,0 +1,192 @@
...
+	cnv := fyne.CurrentApp().Driver().CanvasForObject(c)
+
+	if c.itemHeight == 0 {
+		// set item height to cache
+		c.itemHeight = c.navigableList.CreateItem().MinSize().Height
+	}
+
+	listheight := float32(len(c.Options))*(c.itemHeight+2*theme.Padding()+theme.SeparatorThicknessSize()) + 2*theme.Padding()
+      canvasSize = cnv.Size()
+      entrySize = c.Size()
```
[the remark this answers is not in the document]
metal3d: ho I didn't make the test my self, using the "commit" button without verification, sorry !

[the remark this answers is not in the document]
fpabl0: It is the first time that I do a github suggestion like this, it seems to be dangerous 😅 I won't use it again, at least no without verify it.

fpabl0 (changes requested):
Everything seems to be right now, however I think you should pull the code and formatting because the suggestion I did seems not to be correctly aligned, really sorry! 😞

Also, I am not sure about the file name (`entry-with-completion.go`), shouldn't it be better with underscores instead of dashes and respecting the constructor name? Something like:
```
completion_entry.go
completion_entry_test.go
```
Any thoughts? @andydotxyz

metal3d:
> Everything seems to be correct now, however I think you should pull the code and format it because the suggestion I did seems not to be correctly aligned, really sorry! 

Right, Vim fixed this, no problem

> Also, I am not sure about the file name (`entry-with-completion.go`), shouldn't it be better with underscores instead of dashes and respecting the constructor name? Something like:
> 
> ```
> completion_entry.go
> completion_entry_test.go
> ```

I changed names, I'm OK with this. The name was chosen without any conviction at first 😉 

> Any thoughts? @andydotxyz

Jacalz:
I think completion_entry is good. Matches select_entry naming.

metal3d:
Do not merge right now, I found a bug... I'm creating unit test and fix it

metal3d:
Fixed! There was a problem when the user pressed "Enter", the entry was submitted with the "previous" entry value. There were another problem with the selected id that can be -1.

I fixed both and adder tests.

widget/completion_entry.go
```diff
@@ -0,0 +1,197 @@
...
+	c.ExtendBaseWidget(c)
+	return c
+}
+
+// ShowCompletion displays the completion menu
+func (c *CompletionEntry) ShowCompletion() {
+	if c.pause {
+		return
+	}
+
```
fpabl0, line 31: I think Completion should be hidden if there is no Options to show, otherwise you will see an empty overlay (if the popup menu was previously shown), so adding these lines maybe make sense?:
```go
if c.pause {
	return
}
if len(c.Options) == 0 {
	c.HideCompletion()
	return
}
```

widget/completion_entry.go
```diff
@@ -0,0 +1,197 @@
...
+	if c.popupMenu != nil {
+		c.popupMenu.Resize(c.maxSize())
+		c.popupMenu.Move(c.popUpPos())
+	}
+}
+
+// Prevent the menu to open when the user validate value from the menu.
+func (c *CompletionEntry) setTextFromMenu(s string) {
+	c.pause = true
+	c.Entry.SetText(s)
```
fpabl0, line 107: Maybe the cursor should be shown at the end of the text when the user selects an option?:
```go
...
c.Entry.SetText(s)
rs := []rune(s)
c.Entry.CursorColumn = len(rs)
...
```

widget/completion_entry_test.go
```diff
@@ -0,0 +1,161 @@
...
+func TestCompletionEntry_OnSubmit(t *testing.T) {
+	entry := createEntry()
+	win := test.NewWindow(entry)
+	win.Resize(fyne.NewSize(500, 300))
+	defer win.Close()
+
+	entry.OnSubmitted = func(s string) {
+		entry.HideCompletion()
+		assert.True(t, entry.popupMenu.Hidden)
+	}
```
fpabl0, line 134: I am aware this test works, but maybe it is better to ensure this is called otherwise the assertion could not be performed and the test would be a false positive:
```go
submitCalled := false
entry.OnSubmitted = func(s string) {
	submitCalled = true
	entry.HideCompletion()
	assert.True(t, entry.popupMenu.Hidden)
}
...
win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyReturn})
assert.True(t, submitCalled)
```

widget/completion_entry_test.go
```diff
@@ -0,0 +1,161 @@
...
+		t.Fail()
+	}
+
+	entry.OnChanged = func(s string) {
+		entry.ShowCompletion()
+	}
+	entry.SetText("foo")
+	win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyDown})  // select foofoo
+	win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyEnter}) // OnSubmitted should NOT be called
+}
```
fpabl0, line 161: Same here:
```go
submitCalled := false
entry.OnSubmitted = func(s string) {
	submitCalled = true
}
...
win.Canvas().Focus(entry)
entry.SetText("foo")

win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyDown}) // select foofoo
assert.False(t, submitCalled)
win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyEnter}) // OnSubmitted should NOT be called
assert.False(t, submitCalled)
win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyEnter}) // OnSubmitted should be called
assert.True(t, submitCalled)
```

fpabl0 (commented):
Just some comments :)

widget/completion_entry_test.go
metal3d: I'm not OK with this, this test is made to check an issue where the "OnSumitted" should not be called. Testing if the submission is made is done in the other test.

widget/completion_entry_test.go
metal3d: Like my previous comment, this tests are made to check a known issue, the behavior to test is a "non call"

widget/completion_entry.go
metal3d: I'm not sure about this. I did it before but is it the wanted behavior and is it common ? If yes, it's a simple line of code to add 😄

widget/completion_entry.go
metal3d: I'm not sure about this, I think that the best to do is to leave the user to choose. So it's only needed to hide the completion when the user decide it.

metal3d: Ha right...

widget/completion_entry_test.go
fpabl0: I am confused then, this test is called:
```go
// Test sumbission with opened completion.
func TestCompletionEntry_OnSubmit(t *testing.T) {
...
}
```
Isn't this testing the submission?

widget/completion_entry_test.go
fpabl0: The test name is `TestCompletionEntry_DoubleSubmissionIssue`, but you are only trying to trigger (theoretically) one submit event, shouldn't you call another one and verify if this time everything went fine? 🤔 Anyway, as I said at the beginning, the tests are ok, these are just my thoughts because I am paranoic with false positives 😅, feel free to ignore this if this has no sense.

widget/completion_entry.go
fpabl0: Not sure either. Any thoughts? @Jacalz @andydotxyz

andydotxyz: I think that is up to this widget really... I would expect that makes sense, but I don't think there is a precedent to look at inside the toolkit. Other than `test.Type` I suppose, where we simulate a user typing.

andydotxyz:
> completion_entry.go
> completion_entry_test.go
> Any thoughts? @andydotxyz

I would think that `completionentry.go` and `completionentry_test.go` would best match our style...

fpabl0:
> I would think that completionentry.go and completionentry_test.go would best match our style...

In core repository, we have `select_entry` instead of `selectentry` 🤔. However in fyne-x, we have `filetree` and not `file_tree`, so what would be the best style?

andydotxyz:
Yes we have some of both :(.
I felt that it was better for readability if underscore separated tests, os specifics etc. But not at the cost of overall readability. I think radio_group.go should have been radiogroup.go really, but some are less clear. I guess this and how we split up the tests files should really be part of our design spec.

widget/completion_entry.go
metal3d: OK, I made the changes - this will be here on the next push, thanks 😉

widget/completion_entry_test.go
metal3d: My god, sorry I didn't seen the "submitCalled" variable you added. OK, no problem then.

widget/completion_entry_test.go
metal3d: OK, I actually didn't understand what was the problem for you. I now understand that you want to avoid false positives.

metal3d: The problem is now that the entry loses focus after the first Enter key pressed... This doesn't happen in real life, only inside the test.

metal3d:
Is it better that I close the conversation as soon as I resolve your change request or if I only need to re-request review?

widget/completion_entry_test.go
metal3d: Found... I need to "force" the focus (and I really don't know why, but that makes the job)

widget/completion_entry.go
metal3d: Done https://github.com/fyne-io/fyne-x/pull/11/commits/7504cfc4ffba90b954f425589b6ff5604bded166

widget/completion_entry_test.go
fpabl0: > Found... I need to "force" the focus (and I really don't know why, but that makes the job)

Hmm the above code doesn't work for you? I tested it locally and the test passed 🤔

metal3d: Done https://github.com/fyne-io/fyne-x/pull/11/commits/d5226ec4ef7f385c9ef6207bfa7a68a8ce37d30c

widget/completion_entry_test.go
metal3d: Done https://github.com/fyne-io/fyne-x/pull/11/commits/d5226ec4ef7f385c9ef6207bfa7a68a8ce37d30c

widget/completion_entry_test.go
fpabl0: Hmm I don't think this line is needed:
```go
fyne.CurrentApp().Driver().CanvasForObject(c).Focus(c) // regain focus
```
It was added just because of the test?

widget/completion_entry.go
metal3d: I did changes in  https://github.com/fyne-io/fyne-x/pull/11/commits/4056661a28e28c8488097361680b52a6e59637b8

fpabl0: Great thanks :)

widget/completion_entry_test.go
metal3d: Yes... In test the focus is lost. And without this, the  double submission test fails.

widget/completion_entry.go
fpabl0: @metal3d You should use the length of the runes instead of bytes, as I mentioned above. Because Entry works internally with runes not with bytes

widget/completion_entry_test.go
fpabl0: > Yes... In test the focus is lost. And without this, the double submission test fails.

The test passes without it. I think you forgot one line from my above suggestion:
```go
...
win.Canvas().Focus(entry) // 👈 here you are requesting focus (entry.SetText does not request focus)
entry.SetText("foo")

win.Canvas().Focused().TypedKey(&fyne.KeyEvent{Name: fyne.KeyDown}) // select foofoo
...
```

widget/completion_entry_test.go
fpabl0: > My god, sorry I didn't seen the "submitCalled" variable you added

No worries :)

fpabl0:
>  I guess this and how we split up the tests files should really be part of our design spec.

Yes, agree.

widget/completion_entry.go
metal3d: My bad, I don't know why I did it... Sorry

widget/completion_entry_test.go
metal3d: Here, it says that there is a nil pointer dereference

metal3d: Yes, forgotten... (I should sleep before making changes...)
Fixed here: https://github.com/fyne-io/fyne-x/pull/11/commits/5eeb4701a65a9f5e3ad429ac310caa206af44d4a

metal3d: and commented here https://github.com/fyne-io/fyne-x/pull/11/commits/be3a9faf961816757cd3e680a9300f7775dd6399

widget/completion_entry.go
metal3d: Fixed: https://github.com/fyne-io/fyne-x/pull/11/commits/839d80dedc3e98730d202b8d151b80be71ceb303

fpabl0 (approved):
LGTM! :)

widget/completion_entry_test.go
fpabl0: Great! Thanks :)

widget/completion_entry.go
fpabl0: > My bad, I don't know why I did it... Sorry

No problems :)

andydotxyz:
> Is it better that I close the conversation as soon as I resolve your change request or if I only need to re-request review?

We typically ask the person who opened the conversation to marked it as review, otherwise we loose track of whether a comment was addressed in the eyes of the person reporting it.

metal3d:
> > Is it better that I close the conversation as soon as I resolve your change request or if I only need to re-request review?
> 
> We typically ask the person who opened the conversation to marked it as review, otherwise we loose track of whether a comment was addressed in the eyes of the person reporting it.

That makes sense. I realized this after closing conversation when I fixed the request. Thanks 😉

README.md
andydotxyz: Pretty sure this is missing a return?

widget/completionentry.go
```diff
@@ -0,0 +1,203 @@
...
+// Refresh the list to update the options to display.
+func (c *CompletionEntry) Refresh() {
+	c.Entry.Refresh()
+	if c.navigableList != nil {
+		c.navigableList.SetOptions(c.Options)
+	}
+}
+
+// calculate where the popup should appear
+func (c *CompletionEntry) popUpPos() fyne.Position {
```
andydotxyz, line 71: Please sort the methods alphabetically, with exported first then private, more info at https://github.com/fyne-io/fyne/wiki/Contributing

widget/completionentry.go
```diff
@@ -0,0 +1,203 @@
...
+			n.OnSelected(n.selected)
+		}
+	case fyne.KeyEscape:
+		n.hide()
+	default:
+		n.entry.TypedKey(event)
+
+	}
+}
+func (n *navigableList) TypedRune(r rune) {
```
andydotxyz, line 199: Can you leave a space between methods please?

andydotxyz (changes requested):
Just a few comments on style.
Could you also please update the image so it reflects the fixes for drop-down height?

widget/completionentry.go
metal3d: @andydotxyz My editor uses `goimports` as explained in the contribution guide but the methods are not ordered.
This should happen because the methods you point are private I guess. So I will do it manually...

widget/completionentry.go
metal3d: For TypedRune method it's OK to add a new line, but for the 2 latest, adding a new line is weird IMHO and will remove the alignment of brackets:

``` go
func (n *navigableList) FocusGained() {}
func (n *navigableList) FocusLost()   {}

// after goimports (automatic in my editor)
func (n *navigableList) FocusGained() {}

func (n *navigableList) FocusLost() {}

```

widget/completionentry.go
metal3d: I made changes, can you tell me if it's OK now ?

widget/completionentry.go
metal3d: I did changes, tell me if there are other things to do ;)

metal3d:
> Just a few comments on style.
> Could you also please update the image so it reflects the fixes for drop-down height?

I will 😄  - in a couple of hours

andydotxyz:
I just resolved a merge commit after another addition to fyne-x landed, If you have further commits to land please pull from remote first :).

metal3d:
Ha sorry, forgotten.

metal3d:
Image changed

andydotxyz (changes requested):
Just some formatting that's still outstanding

metal3d:
> Just some formatting that's still outstanding

done

widget/completionentry.go
andydotxyz: Thanks. Sadly goimports does not fix the ordering of methods. 
We're working on it ;)

andydotxyz (approved):
Excellent thanks

andydotxyz:
@Jacalz Do you want to confirm the revisions as well?

Jacalz:
> @Jacalz Do you want to confirm the revisions as well?

No need. Land it if you want.

andydotxyz:
Excellent, landed

andydotxyz:
Congratulations and thanks for the component @metal3d

Jacalz:
Indeed. Very well done.

metal3d:
Thanks to you!

A pity that I spent so much time respecting the guidelines. But now that I know your rules, the next one should be easier to validate ;)

Issue #9, opened 2021-02-18
Conflicting package names


tehsphinx:
As a user of the fyne framework who wants to use some widgets or features from `fyne-x` I will have to deal with package naming conflicts a lot. Both `fyne` and `fyne-x` modules have a package `widgets` and a package `layout`.

andydotxyz:
We have mirrored the names for clarity, and with Go import naming this should not be a problem.
```go
import x-widgets "fyne.io/x/fyne/widgets`
```
What would your suggestion be?

tehsphinx:
I'd probably try to avoid the same names as it is hopefully be used together a lot. It could actually be called `xwidgets` and `xlayout`.

Then autoimport would work in editors as the editor would know what I want to import. Especially if `fyne/widgets` is already imported, I can't type `widget.SomeXWidget` and I can't type `xwidget.SomeXWidget`. The first one doesn't work because there is another widget package imported already and SomeXWidget just doesn't exist there. The second does not work because there is no such package since I would need to set that alias first.

andydotxyz:
> Then autoimport would work in editors as the editor would know what I want to import.

It works for me, the IDE indexes all packages.

> I can't type widget.SomeXWidget and I can't type xwidget.SomeXWidget.

If you specify a package name then indeed it won't find elements that are not in that package. Search instead for "SomeXWidget" and it will find the correct package. In GoLand it will even import and name appropriately the secondary package.

andydotxyz:
Also for background, we were following the golang.org/x/ naming.
I.e. `net` is a builtin import and `golang.org/x/net` is  where they extend the capabilities.

fpabl0:
> net is a builtin import and golang.org/x/net is where they extend the capabilities.

That's a good example, however they don't have the same problem as here, because `golang.org/x/net` don't expose any type, function or constant with `net` namespace, it has divided in more packages that do not cause naming conflict with `net`.

I think @tehsphinx's proposal is good (indeed I was about to write an issue with the same title 😅). 

>  In GoLand it will even import and name appropriately the secondary package.

Unfortunately, not many devs can get GoLand as it is not free, I think most of them (me included) are going to use VSCode.

Or I am not sure, but `firebase` go sdk use something that automatically put the name of the package as an alias:
```go
firebase "firebase.google.com/go"
```
Looking at the source code, they have:
```go
package firebase // import "firebase.google.com/go"
```

So, maybe `fyne-x` could have a similiar approach?
Something like:
```go
[...]

PR #12, opened 2021-04-26
Add a password validator

Jacalz:
A validator that I have used in [Sparta](https://github.com/Jacalz/sparta). I hope that it can be useful for more people. Was unsure if this should go in regular fyne or not, but decided on placing it here due to the new import.

### DIFF
--- a/data/validation/password.go
+++ b/data/validation/password.go
@@ -2,9 +2,8 @@
 package validation // import "fyne.io/x/fyne/data/validation"
 
 import (
-	gpv "github.com/wagslane/go-password-validator"
-
 	"fyne.io/fyne/v2"
+	gpv "github.com/wagslane/go-password-validator"
 )
 
 // NewPassword returns a new validator for validating passwords.
--- a/data/validation/password.go
+++ b/data/validation/password.go
@@ -2,8 +2,9 @@
 package validation // import "fyne.io/x/fyne/data/validation"
 
 import (
-	"fyne.io/fyne/v2"
 	gpv "github.com/wagslane/go-password-validator"
+
+	"fyne.io/fyne/v2"
 )
 
 // NewPassword returns a new validator for validating passwords.
--- a/data/validation/password.go
+++ b/data/validation/password.go
@@ -1,0 +1,18 @@
+// Package validation provides validation for data inside widgets
+package validation // import "fyne.io/x/fyne/data/validation"
+
+import (
+	"fyne.io/fyne/v2"
+	gpv "github.com/wagslane/go-password-validator"
+)
+
+// NewPassword returns a new validator for validating passwords.
+// Validate returns nil if the password entropy is greater than or equal
+// to the minimum entropy. If not, an error is returned that explains
+// how the password can be strengthened. Advice on entropy value:
+// https://github.com/wagslane/go-password-validator/tree/main#what-entropy-value-should-i-use
+func NewPassword(minEntropy float64) fyne.StringValidator {
+	return func(text string) error {
+		return gpv.Validate(text, minEntropy)
+	}
+}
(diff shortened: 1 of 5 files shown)


andydotxyz (commented):
This is a pretty cool addition, but I found the single number parameter to be confusing. I read the docs and understand how it works - but I wondered if perhaps some constants could help to illustrate basic/high/extreme entropy levels for those who don’t want to read abs decide on a number for their code?

Jacalz:
I had a similar idea initially as well but decided against doing so because I was afraid that it would cause more confusion. I feel like ost people, being unsure, would take the lowest option and the max numbers would give a false sense of security. That was why I decided to just stick the link in there instead.

stuartmscott:
The devs who use this will need to read the docs to figure out the right number for their app, I don't think Fyne should create "security buckets" as the landscape continuously changes and they will soon be out-of-date/unsafe.

I can see this being useful to some devs, although I think Fyne-X might be a better place so that people who don't need this can avoid the extra dependency.

Jacalz:
> I can see this being useful to some devs, although I think Fyne-X might be a better place so that people who don't need this can avoid the extra dependency.

This is already in fyne-x ;)

fpabl0:
This is a great addition :). However I think folders should be independent modules (like I mentioned in #9), in that way users that only need extension widgets don't have to depend on dependencies used in `data/validation` for example.

stuartmscott:
> This PR is already in fyne-x ;)

Oops, not enough coffee :)

Jacalz:
> This is a great addition :). However I think folders should be independent modules (like I mentioned in #9), in that way users that only need extension widgets don't have to depend on dependencies used in `data/validation` for example.

Have we gotten an agreement on that? I personally think it is a great idea, but would like to hear what other people think about it before take that route.

fpabl0:
> Have we gotten an agreement on that?

Not yet. That is something that needs to be discussed maybe in a new issue? 🤔

Jacalz:
We could perhaps discuss it on the call later this week?

fpabl0:
> We could perhaps discuss it on the call later this week?

Sounds like a good idea :)

andydotxyz:
> We could perhaps discuss it on the call later this week?

Cool, if you would like it discussed please add to the agenda. Recently we've been spending around 2/3 of the time on non-agenda items so I want to get it tidied up.

fpabl0:
> Cool, if you would like it discussed please add to the agenda. 

Added.

andydotxyz:
Should now we structure fyne-x modules be holding up this contribution?

data/validation/password.go
```diff
@@ -0,0 +1,18 @@
+// Package validation provides validation for data inside widgets
+package validation // import "fyne.io/x/fyne/data/validation"
+
+import (
+	"fyne.io/fyne/v2"
+	gpv "github.com/wagslane/go-password-validator"
```
andydotxyz, line 6: I think  this is missing a newline between the import groups.
Maybe we have not got the full suite of checks running on this repo yet?

Jacalz: The workflow is a few versions behind, but still fully functional. My editor formats stuff on save as well, it did not catch it and running goimports locally also does not correct it. I think it might be the explicit import name that makes it wonky. Either way, fixed now.

Jacalz:
I don't think so. Isn't it smoother to re-shuffle everything at the same time?

data/validation/password.go
andydotxyz: This is not fixed now, sorry. As we are in an external project (fyne-x) the fyne.io imports come before github.com due to alphabetical ordering of external imports

Jacalz: Good point. Thinking about this again, there probably should not be a newline between them, as we are not in the same repository.

andydotxyz (approved):

PR #13, opened 2021-05-08
update README to reference the new hex widget

charlesdaniels:
Edit: this PR now also moves the new hex widget into the main `widget` package, per @andydotxyz.


charlesdaniels:
Looks like maybe go.mod needs updated, but I don't think that's related to this PR, since there were no code changes. Did `app` change upstream or something? Should I fix this as part of this PR, or does that need to be done separately.

andydotxyz (changes requested):
Thanks for all this, but can the images be put into img/ please?

charlesdaniels:
> Thanks for all this, but can the images be put into img/ please?

Done in b67a3c2.

README.md
```diff
@@ -51,6 +51,25 @@ tree.Sorter = func(u1, u2 fyne.URI) bool {
 </p>
 
 
+### 7-Segment ("Hex") Display
+
+A skeuomorphic widget simulating a 7-segment "hex" display. Support setting
```
Jacalz, line 56: Shouldn’t it be “Supports” with an “s” at the end?

charlesdaniels: 6411fe6

andydotxyz (approved):

PR #16, opened 2021-08-31
SliderSwitch, SingleRuneEntry, and SegmentedEntry widgets

tinyzimmer:
See #15 for more information


widget/segmentedentry.go
```diff
@@ -0,0 +1,230 @@
...
+		allowedRunes: allowedRunes,
+		length:       length,
+		entries:      make([]*SingleRuneEntry, length),
+		selected:     0,
+	}
+	input.ExtendBaseWidget(input)
+	return input
+}
+
+// GetValue returns the value of all the entries in the grid combined.
```
andydotxyz, line 50: I don't think the `Get` prefix is normally used

widget/segmentedentry.go
```diff
@@ -0,0 +1,230 @@
...
+				}
+				g.FocusIndex(this + 1)
+			}
+		}
+	}
+
+	return &segmentedEntryRenderer{input: g}
+}
+
+func (g *SegmentedEntry) objects() []fyne.CanvasObject {
```
andydotxyz, line 149: Just a suggestion, but it seems like this would be cleaner in the Objects() function of the renderer to save this additional method.

widget/singleruneentry.go
```diff
@@ -0,0 +1,124 @@
...
+	//  Backspace
+	//    - Called with the key if the length of this entry is
+	//      already zero.
+	//
+	//  TextAdded
+	//    - Called with key set to a nil value
+	//      TODO: Probably have a separate OnChanged or fall up to
+	//      the widget.Entry implementation. Or declare new typed
+	//      constants.
+	OnNavigate func(key *fyne.KeyEvent)
```
andydotxyz, line 49: I don't think this method is needed - if developers have to filter on the key they would be just as well to extend and use KeyTyped like with a regular entry.

widget/sliderswitch.go
```diff
@@ -0,0 +1,58 @@
...
+	widget.Slider
+
+	// OnToggle is called AFTER the switch has changed state.
+	// Bool is true if the switch is in the "ON" state.
+	OnToggle func(bool)
+	// If set, the slider will be moved to an intermediate state,
+	// the function called, and then moved to its final state after
+	// the function returns. Bool is true if the switch is transitioning
+	// to the "ON" state.
+	OnTransition func(bool)
```
andydotxyz, line 22: I don't understand what transitioning is and/or why value 1 between 0 and 2...
For one thing how can you possibly know that 0->1 then follows with 2, it could be 0->1->0.

widget/sliderswitch.go
```diff
@@ -0,0 +1,58 @@
...
+var _ fyne.Tappable = (*SliderSwitch)(nil)
+
+// SliderSwitch is an extension of a slider that switches
+// between two values when clicked.
+type SliderSwitch struct {
+	widget.Slider
+
+	// OnToggle is called AFTER the switch has changed state.
+	// Bool is true if the switch is in the "ON" state.
+	OnToggle func(bool)
```
andydotxyz, line 17: As above, try `ed` suffix for a callback that happens AFTER.

andydotxyz (changes requested):

widget/singleruneentry.go
tinyzimmer: I must have missed `KeyTyped` - I'll look into that because it sounds like it would simplify things.

andydotxyz: Sorry, I typeoed - I meant TypedKey, like you already use in the code

widget/sliderswitch.go
andydotxyz: Don't folk normally set the state and then disable to show that it is not available then re-enable when completed?
I wouldn't be exposing transitioning in this way - users can slide to the intermediate position. Also taking this model means the UI will seem to "stick" part way through, looking unresponsive - but it can still be interacted with. It feels untidy I guess.

tinyzimmer: Fair points

Jacalz (changes requested):
Thanks for all of this. A few quick notes from looking at the code :)

widget/segmentedentry.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// A function to run when the grid is full
+	OnCompletion func(string)
+	// Text to use to separate fields. It will be
+	// forced to monospace to ensure consistent
+	// sizing.
+	Delimiter string
+	// Optionally specify which fields should have
+	// a separator after them.
+	DelimitAt []int
```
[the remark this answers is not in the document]
Jacalz: Never mind. Might have been clearer if you had all the examples in the description here.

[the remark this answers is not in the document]
tinyzimmer: I might be understanding you wrong 😛 . But how could that translate to indices to place delimiters at?

widget/segmentedentry.go
```diff
@@ -0,0 +1,230 @@
...
+	input.ExtendBaseWidget(input)
+	return input
+}
+
+// GetValue returns the value of all the entries in the grid combined.
+func (g *SegmentedEntry) GetValue() string {
+	var out string
+	for _, e := range g.entries {
+		out += strings.TrimSpace(e.Text)
+	}
```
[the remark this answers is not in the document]
tinyzimmer: The reason I didn't at first was the need for error checking on the [Write methods](https://pkg.go.dev/strings#Builder). I was worried it would piss off the linter in a function that under normal circumstances wouldn't need to worry about error checking.

widget/segmentedentry.go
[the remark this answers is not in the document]
Jacalz: It was a poor idea from my side. Forget it ;)

widget/segmentedentry.go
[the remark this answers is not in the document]
Jacalz: The write methods never returns any errors, check the  documentation. No need to handle them ;)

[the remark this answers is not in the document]
tinyzimmer: But would the linter be as forgiving?

widget/singleruneentry.go
tinyzimmer: I'm trying other ways of implementing this, and I'm wondering if it is better for it to not be an exported widget. The OnNavigate serves the focusing functionality in the `SegmentedEntry`. I'm worried removing it here will just lead to more complexity there. Though the widget _could_ see use for other forms where a user just wants to accept a single rune, it wasn't the original intention.

Thoughts?

tinyzimmer: Though on second thought, the aforementioned user wanting an entry which accepts a single rune...they could use a SegmentedEntry with length set to 1.

tinyzimmer:
I probably won't have time to finish these PRs 😞 . But I'll chime back if that ever changes.

Jacalz:
Oh no, that's unfortunate to hear. Is this something you would would like help with landing? I could potentially work on getting this the last meters to landing if it's something you'd be happy with.

tinyzimmer:
You are more than welcome to re-open and push it over the edge, and I can try to help if needed. I've just been pretty swamped with the day-job and haven't had much energy on the weekends.

Jacalz:
Understandable. I will try to se what I can do. I'm in a relatively similar situation with university but I will make sure to put this on my list :)

Jacalz:
Just reopening to make sure that it isn’t forgotten about. This is something that I think would be useful in one way or another. If you don’t want to continue working in this @tinyzimmer, I might pick it up sometime like I said before :)

Jacalz:
I love where this PR is headed and I would definitely want to complete it. However, I doubt if I'll have time to complete it as I need to focus on other stuff. I'll close this and open an issue to let someone else continue the implementation :)

Jacalz:
@tinyzimmer Is of course also very much welcome to continue it if he wants to. If not, I'll open an issue to keep track of the idea 💡

PR #19, opened 2021-10-06
Updates to the project and deps

Jacalz:
This updates Fyne to 2.1.0 and fixes that the modules file incorrectly mentioned Go 1.15 as lowest version. I also updated testify for bug fixes and an optimization (see https://github.com/stretchr/testify/pull/933#issuecomment-620217067). Lastly, I added a .gitignore file based on the one in fyne-io/fyne.

A bit much in one PR perhaps but I did not see it as fair for me to get 5 PRs for hacktoberfest for such small changes. Every little change is separated into their own commits,


Jacalz:
Looks like this is going to be blocked until https://github.com/fyne-io/fyne-x/pull/18 lands. The updated testify requires Go 1.13 or later, it seems.

andydotxyz:
Nice fix.
I think you're right it needs to rebase on (just approved) #18

PR #25, opened 2021-11-08
Add map widget

andydotxyz:
Pretty simple, offers zoom and pan in integer increments.
Lots more work could be done but it feels ready for a first PR / discussion.


widget/map.go
```diff
@@ -0,0 +1,148 @@
...
+
+	cl := &http.Client{}
+	for x := firstTileX; (x-firstTileX)*tileSize <= w+tileSize/2; x++ {
+		for y := firstTileY; (y-firstTileY)*tileSize <= h+tileSize/2; y++ {
+			if x < 0 || y < 0 || x >= int(count) || y >= int(count) {
+				continue
+			}
+
+			u := fmt.Sprintf("https://tile.openstreetmap.org/%d/%d/%d.png", m.zoom, x, y)
+			req, err := http.NewRequest("GET", u, nil)
```
Bluebugs, line 121: I am wondering how difficult it would be to introduce some caching here or if there is already some.

Bluebugs (approved):
Pretty simple and quickly done map widget, neat!

I wonder if we could move maybe some improvement to core Fyne, as the ability to download and display a remote image directly from a url would be useful to more than just a map widget.

Jacalz (commented):
This is looking really cool. Nice work!
It would be very cool to add touch and drag plus scroll to zoom. I think that would make for a very compelling first version of a map widget :)

andydotxyz:
> It would be very cool to add touch and drag plus scroll to zoom. I think that would make for a very compelling first version of a map widget :)

Is that going to block this PR? I have tidied and addressed the other PR comments, wasn't sure if this was needed too, or could it be another set of work?

widget/map.go
andydotxyz: Thanks, added - I should have done this already as it is part of the openstreetmap requirements.

Jacalz:
> > It would be very cool to add touch and drag plus scroll to zoom. I think that would make for a very compelling first version of a map widget :)
> 
> 
> 
> Is that going to block this PR? I have tidied and addressed the other PR comments, wasn't sure if this was needed too, or could it be another set of work?

Just a suggestion. I'll not let it block anything for now.

widget/map.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	midTileX := (w - tileSize*2) / 2
+	midTileY := (h - tileSize*2) / 2
+	if m.zoom == 0 {
+		midTileX += tileSize / 2
+		midTileY += tileSize / 2
+	}
+
+	count := math.Pow(2, float64(m.zoom))
```
Jacalz, line 121: I think it kind of is both clearer and faster to just do this instead:
```suggestion
	count := float64(m.zoom) * float64(m.zoom)
```

Jacalz (changes requested):
I just had a very minor suggestion. Apart from that, I think the map image for the README, might need and update.

widget/map.go
andydotxyz: But 2^zoom is not the same as zoom^2

Jacalz: Gosh. How did I not see that it was 2^x? Sorry for the fuss.

andydotxyz:
It still says you are requesting changes @Jacalz but i cannot see what they are...

widget/map.go
Jacalz: You could actually use a binary shift instead and do `1 << m.zoom`. From my simple benchmarks, it seems to be about 60 times faster.

Jacalz (commented):
You are right, sorry. I added another comment to my last inline review. I'll approve after that has been answered :)

stuartmscott:
Linking for posterity; https://gophers.slack.com/archives/CK5U4BU87/p1637018319438000

widget/map.go
andydotxyz: Thanks, done

Jacalz (approved):
This is so slick. Nice work! :)

PR #27, opened 2021-12-09
Add support for a mqtt string binding.

Bluebugs:
As you can get message on multiple topic on the same Mqtt connection, that connection has to be established ahead of getting binding to a topic. A small example to test this binding:
```
	var broker = "broker.emqx.io"

	var port = 1883
	opts := mqtt.NewClientOptions()
	opts.AddBroker(fmt.Sprintf("tcp://%s:%d", broker, port))
	opts.SetClientID("fyne_demo")
	opts.SetUsername("emqx")
	opts.SetPassword("public")
	opts.OnConnect = connectHandler
	opts.OnConnectionLost = connectLostHandler
	client := mqtt.NewClient(opts)
	token := client.Connect()
	token.Wait()

	s, _ := xbinding.NewMqttString(client, "fyne.io/x/string")
```

You can then watch change with:
`mosquitto_sub -h broker.emqx.io -t fyne.io/x/string`
And make change with:
`mosquitto_pub -h broker.emqx.io -t fyne.io/x/string -m "youhou"`


README.md
```diff
@@ -134,6 +134,14 @@ The code above uses a test web sockets server from "PieSocket", you can run the
 and go to [their test page](https://www.piesocket.com/websocket-tester) to send messages.
 The widget will automatically update to the latest data sent through the socket.
 
+### MqttString
+
+A `MqttString` binding creates a `String` data binding to the specified _topic_ associated with
+the specified **MQTT** client connection. Each time a message is received the value will be converted
+to a `string` and set on the binding. Each time the value is edited, it will be sent back over
+**MQTT** on the specified _topic_. It is also a `Closable` so you should be sure to call `Close`
```
andydotxyz, line 142: I think this should be `Closer` not `Closable`.

README.md
```diff
@@ -134,6 +134,14 @@ The code above uses a test web sockets server from "PieSocket", you can run the
...
+once you are completed using it to disconnect the _topic_ handler from the **MQTT** client connection.
+
```
andydotxyz, line 144: Would it be possible to add a little usage info or demo in here, just make it a bit clearer how to use?

data/binding/mqttstring.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}()
+
+	ret := &mqttString{String: s, conn: conn, topic: topic}
+
+	return ret, nil
+}
+
+func (s *mqttString) Set(val string) error {
+	token := s.conn.Publish(s.topic, 0, false, val)
```
andydotxyz, line 41: Should this also call `s.String.Set(val)`?
Doing so would mean any local bindings would refresh sooner, rather than waiting for the MQTT round trip...

andydotxyz (changes requested):
Cool, just some thoughts inline

data/binding/mqttstring.go
Bluebugs: It is something I wondered about. There is no guarantee that the value sent over will actually be the same that come back, that's why I settle on not updating locally.

data/binding/mqttstring.go
```diff
@@ -0,0 +1,68 @@
...
+	err   error
+}
+
+// NewMqttString returns a `String` binding to a web socket server specified as `url`.
+// The resulting string will be set to the content of the latest message sent through the socket.
+// You should also call `Close()` on the binding once you are done to free the connection.
+func NewMqttString(conn mqtt.Client, topic string) (StringCloser, error) {
+	ret := &mqttString{String: binding.NewString(), conn: conn, topic: topic}
+
+	go func() {
```
stuartmscott, line 23: Can this be synchronous so the dev has control eg. so they can show an infinite progress bar?

data/binding/mqttstring.go
```diff
@@ -0,0 +1,68 @@
...
+		token := conn.Subscribe(topic, 1, func(c mqtt.Client, m mqtt.Message) {
+			ret.String.Set(string(m.Payload()))
+		})
+
+		token.Wait()
+
+		if token.Error() != nil {
+			ret.err = token.Error()
+		}
```
stuartmscott, line 32: No need to call the method twice
```suggestion
		ret.err = token.Error()
```

data/binding/mqttstring.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (s *mqttString) Set(val string) error {
+	token := s.conn.Publish(s.topic, 0, false, val)
+
+	token.Wait()
+	if token.Error() != nil {
+		s.err = token.Error()
+		return token.Error()
+	}
```
stuartmscott, line 45: No need to call the method thrice
```suggestion
	if err := token.Error(); err != nil {
		s.err = err
		return err
	}
```

data/binding/mqttstring.go
Bluebugs: Indeed, I forgot about it.

data/binding/closers.go
```diff
@@ -0,0 +1,15 @@
+// Package binding provides extended sources of data binding.
```
andydotxyz, line 1: We should only include the package doc in 1 file

README.md
```diff
@@ -134,6 +134,26 @@ The code above uses a test web sockets server from "PieSocket", you can run the
...
+```go
+opts := mqtt.NewClientOptions()
+opts.AddBroker("tcp://broker.emqx.io:1883")
+opts.SetClientID("fyne_demo")
+client := mqtt.NewClient(opts)
+
+token := client.Connect()
+token.Wait()
```
stuartmscott, line 152: Many people will probably copy/paste this so consider adding the error handling
```suggestion
token.Wait()
if err := token.Error(); err != nil {
    // Handle
}
```

andydotxyz (changes requested):
Looks great, just one thing I noticed that I missed before sorry - package docs should appear only once.

data/binding/mqttstring.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type mqttString struct {
+	binding.String
+	conn  mqtt.Client
+	topic string
+	err   error
+}
+
+// NewMqttString returns a `String` binding to a web socket server specified as `url`.
```
stuartmscott, line 17: > web socket server specified as `url`.

Copy/paste?

data/binding/mqttstring.go
```diff
@@ -0,0 +1,65 @@
...
+func (s *mqttString) Set(val string) error {
+	token := s.conn.Publish(s.topic, 0, false, val)
+
+	token.Wait()
+	if s.err = token.Error(); s.err != nil {
+		return s.err
+	}
+	s.err = nil
+
+	return nil
```
stuartmscott, line 45: Can be simplified
```suggestion
	s.err = token.Error()
	return s.err
```

stuartmscott (approved):

stuartmscott:
Nice work @Bluebugs, btw I kicked off the checks for you - apparently they don't run automatically for first-time contributors to this repo 😕

andydotxyz (approved):
Cool thanks :)

PR #28, opened 2021-12-09
Report errors that are encountered

andydotxyz:
A little refactoring as well, but the major change is to report errors instead of printing.

### DIFF
--- a/data/binding/webstring.go
+++ b/data/binding/webstring.go
@@ -57,7 +57,7 @@ func (s *webSocketString) Get() (string, error) {
 func (s *webSocketString) readMessages() {
 	for {
 		_, p, err := s.conn.ReadMessage()
-		s.prev = err // if no error we clear the state
+		s.prev = err    // if no error we clear the state
 		if err != nil { // permanent (could be connection closed)
 			return
 		}
--- a/data/binding/webstring.go
+++ b/data/binding/webstring.go
@@ -3,10 +3,8 @@ package binding
 
 import (
 	"io"
-	"log"
 	"net/http"
 
-	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/data/binding"
 
 	"github.com/gorilla/websocket"
@@ -22,6 +20,7 @@ type StringCloser interface {
 type webSocketString struct {
 	binding.String
 	conn *websocket.Conn
+	prev error
 }
 
 // NewWebSocketString returns a `String` binding to a web socket server specified as `url`.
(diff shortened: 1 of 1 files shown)


data/binding/webstring.go
```diff
@@ -59,3 +45,28 @@ func (s *webSocketString) Close() error {
...
+func (s *webSocketString) readMessages() {
+	for {
+		_, p, err := s.conn.ReadMessage()
+		if err != nil { // permanent (could be connection closed)
+			s.prev = err
+			return
+		}
+
+		err = s.Set(string(p))
+		if err != nil { // not permanent
```
stuartmscott, line 68: Seems odd that a subsequent Get would return an error from Set - this error would normally be from a later binding, eg converter.

Maybe I'm miss understanding the use cases; 1) dev creates websocket binding and attaches it to a label, websocket fails to connect so dev must call Get to see what the issue was? 2) websocket connects and receives a message, but binding fails to set the string, so again dev must call Get to see what the issue was? Worse is that either error can occur at any time, so the dev would need to poll Get :(

data/binding/webstring.go
```diff
@@ -34,21 +33,8 @@ func NewWebSocketString(url string) (StringCloser, error) {
...
-				return
-			}
-
-			err = s.Set(string(p))
-			if err != nil {
-				fyne.LogError("Failed to set string from web socket", err)
-			}
-		}
-	}()
 	ret := &webSocketString{String: s, conn: conn}
```
stuartmscott, line 36: nit: consider inlining `s`

data/binding/webstring.go
stuartmscott: I think it might be nicer to be able to set a callback `OnError` so the dev can explicitly handle failures, rather than this implicit behaviour.

andydotxyz: Sorry I missed that this error can never occur, so I now ignore it with a comment

andydotxyz: > 1) dev creates websocket binding and attaches it to a label, websocket fails to connect so dev must call Get to see what the issue was?

If it fails to connect the error is returned from the constructor, no need to call into Get - those are for messages happening later.

data/binding/webstring.go
```diff
@@ -59,3 +44,24 @@ func (s *webSocketString) Close() error {
 
 	return s.conn.Close()
 }
+
+func (s *webSocketString) Get() (string, error) {
+	if s.prev != nil {
+		err := s.prev
+		return "", err
```
stuartmscott, line 51:
```suggestion
		return "", s.prev
```

stuartmscott: or if you don't want to access the field twice;

```go
if e := s.prev; e != nil {
    return "", e
}
```

andydotxyz: Nice, thanks

PR #30, opened 2021-12-20
Add a JSON data binding connected to a string data binding.

Bluebugs:


Bluebugs:
This data binding can be used in combination with any binding.String, but it is really useful with the mqtt binding as most protocol are JSON+MQTT or WebSocket as they also use JSON really frequently.

go.sum
```diff
@@ -19,27 +24,33 @@ github.com/go-gl/gl v0.0.0-20210813123233-e4099ee2221f/go.mod h1:wjpnOv6ONl2SuJS
 github.com/go-gl/glfw/v3.3/glfw v0.0.0-20210410170116-ea3d685f79fb h1:T6gaWBvRzJjuOrdCtg8fXXjKai2xSDqWTcKFUPuw8Tw=
 github.com/go-gl/glfw/v3.3/glfw v0.0.0-20210410170116-ea3d685f79fb/go.mod h1:tQ2UAYgL5IevRw8kRxooKSPJfGvJ9fJQFa0TUsXzTg8=
 github.com/go-ole/go-ole v1.2.5/go.mod h1:pprOEPIfldk/42T2oK7lQ4v4JSDwmV0As9GaiUsvbm0=
+github.com/go-sql-driver/mysql v1.4.0/go.mod h1:zAC/RDZ24gD3HViQzih4MyKcchzm+sOG5ZlKdlhCg5w=
```
andydotxyz, line 27: Surely this does not need mysql, gopherjs etc?
Are you missing a "go mod tidy" recently or are these dependencies of the json value module?

data/binding/json.go
```diff
@@ -0,0 +1,329 @@
...
+// JSONValue supports binding a jsonvalue.V{}
+type JSONValue interface {
+	binding.DataItem
+	sync.Locker
+	RLock()
+	RUnlock()
+
+	Get() (*jsonvalue.V, error)
+	Set(*jsonvalue.V) error
+	GetItemString(firstParam interface{}, params ...interface{}) (binding.String, error)
```
andydotxyz, line 21: Should we support more (almost the complete?) list of standard bound types here? It seems like `Int` is missing at least.
Maybe Rune does not make sense, I don't know JSON very well

andydotxyz (changes requested):
One more change I noticed would be good (`Int`).
Also I think "go mod tidy" has not been run recently looking at the go.sum additions.

data/binding/json.go
```diff
@@ -0,0 +1,329 @@
...
+	generic childJSON
+}
+
+var (
+	errWrongType = errors.New("wrong type provided")
+)
+
+// NewJSONFromDataString return a data binding to a `jsonvalue.V{}` synchronized with the `String`
+// binding used to create the new binding.
+func NewJSONFromDataString(data binding.String) (JSONValue, error) {
```
andydotxyz, line 66: I wonder if the word `Data` in here is not needed? The Input is a `String`

andydotxyz:
It would be nice if we could find a way to not expose the internal locking in the public API.

data/binding/json.go
Bluebugs: JSON only has three data type: string, float and boolean which is why I got with those three. Any other type, would technically be a conversion, so I left them out since we have all the conversion in fyne itself.

andydotxyz: I don't think we have a Float<->Int conversion though, so this is going to be a limitation

andydotxyz: I think JSON number is either int or float isn't it? I.e. I can store "1" as a valid number.

data/binding/json.go
Bluebugs: I think that's a good suggestion, I have dropped Data from the name.

Bluebugs:
> It would be nice if we could find a way to not expose the internal locking in the public API.

I really don't know how to do that part. In C++, you could do some magic using a smart pointer and having a const Get and a non const Get which would allow to use RW lock. But I do not know how to do something equivalent in Go and if it makes sense to do something like that in Go. If anyone has any suggestion, let me know.

andydotxyz:
> I really don't know how to do that part. In C++, you could do some magic using a smart pointer and having a const Get and a non const Get which would allow to use RW lock. But I do not know how to do something equivalent in Go and if it makes sense to do something like that in Go. If anyone has any suggestion, let me know.

Could it make sense to not export them just now, and explore how to support custom additions to this in a future version?
That way we are not adding public API that may change, instead holding it private until we have more knowledge / use-case.

Bluebugs:
> > I really don't know how to do that part. In C++, you could do some magic using a smart pointer and having a const Get and a non const Get which would allow to use RW lock. But I do not know how to do something equivalent in Go and if it makes sense to do something like that in Go. If anyone has any suggestion, let me know.
> 
> Could it make sense to not export them just now, and explore how to support custom additions to this in a future version? That way we are not adding public API that may change, instead holding it private until we have more knowledge / use-case.

That would mean making Set,Get also private for the moment. If that's ok for you, that sounds like ok to me.

Bluebugs:
> > Could it make sense to not export them just now, and explore how to support custom additions to this in a future version? That way we are not adding public API that may change, instead holding it private until we have more knowledge / use-case.
> 
> That would mean making Set,Get also private for the moment. If that's ok for you, that sounds like ok to me.

Actually I would still need to export Get to be useful. This is used in my experience to detect when a full valid json object has been received. I guess just Set,Lock,Unlock then.

andydotxyz:
> That would mean making Set,Get also private for the moment. If that's ok for you, that sounds like ok to me.

Sorry I don't quite follow here. Is it not possible to expose `Get` and `Set` without the public locking mechanism?
Bindings should be able to manage concurrency internally whilst exposing those accessory methods.

Bluebugs:
> > That would mean making Set,Get also private for the moment. If that's ok for you, that sounds like ok to me.
> 
> Sorry I don't quite follow here. Is it not possible to expose `Get` and `Set` without the public locking mechanism? Bindings should be able to manage concurrency internally whilst exposing those accessor methods.

Get can be made to return a copy protecting the value that the data binding hold, but when you want to Set a property, you need to Get and then Set without anything modifying the data in between or you end up with a race condition. So when you do a Set, you need to block all other Set for sure which require some locking mechanism, and if you don't expose a Lock mechanism from the beginning and return a copy, we will only be able to implement a Mutex interface, not a RWMutex.

If we do want to handle the locking internally, it means we need to expose a Get/Set that does not expose the jsonvalue, but only sub field. In some way duplicating the jsonvalue API I think.

andydotxyz:
Users of this api should not be expecting to call Get before Set, nor to be locking for either access.
Maybe the JSONValue API is complicating this or maybe I am missing the point?
Yes I think that locking should be handled internally

Bluebugs:
I have been thinking about this and I can not see a way to make a the Get/Set thread safe. I think the move forward is for the binding to not have Lock interface and Get/Set exposed. Only the children data binding accessor would allow access to the data. The impact on not being able to directly extend the JSON data binding doesn't seems to be a big problem as the base type are all covered and post conversion is still possible.

andydotxyz:
> I have been thinking about this and I can not see a way to make a the Get/Set thread safe.

This is unfortunate, what is the main blocker? I still don't think I quite understand the problem.

> The impact on not being able to directly extend the JSON data binding doesn't seems to be a big problem as the base type are all covered and post conversion is still possible.

Yeah cool

Bluebugs:
The JSON value is a complex structure with multiple way to interact with it in a nested way. There is no atomic way to modify it or make sure that it content doesn't change between a read and a write to it. You could check that a children is an array and before you modify it someone else would have made it a number and things would fail without you having any mean to prevent it.

data/binding/json.go
```diff
@@ -0,0 +1,327 @@
...
+	self   binding.Untyped
+	rwlock sync.RWMutex
+	source binding.String
+	last   string
+	err    error
+}
+
+// Internal type for the children accessor
+type childJSON struct {
+	source JSONValue
```
andydotxyz, line 35: If this was of type `*databoundJSON` you could lose a lot of casts required for the `get()` and `set(...)` calls.

andydotxyz:
> The JSON value is a complex structure with multiple way to interact with it in a nested way. There is no atomic way to modify it or make sure that it content doesn't change between a read and a write to it. You could check that a children is an array and before you modify it someone else would have made it a number and things would fail without you having any mean to prevent it.

Oh, I completely missed that you were returning the embedded `jsonvalue.V`, sorry. Probably no big deal to omit it after all. Let's not expose that `V` as part of our API :)

data/binding/json.go
```diff
@@ -0,0 +1,331 @@
...
+type childJSONBool struct {
+	binding.Bool
+	generic childJSON
+}
+
+var (
+	errWrongType = errors.New("wrong type provided")
+)
+
+// NewJSONFromString return a data binding to a `jsonvalue.V{}` synchronized with the `String`
```
andydotxyz, line 61: As this is not part of the public API I think it would be better to remove it, using something like "binding to a JSON structure" I guess?
(Appears in 1 other place as well)

data/binding/json.go
```diff
@@ -0,0 +1,331 @@
...
+import (
+	"errors"
+	"sync"
+
+	"fyne.io/fyne/v2/data/binding"
+
+	jsonvalue "github.com/Andrew-M-C/go.jsonvalue"
+)
+
+// JSONValue supports binding a jsonvalue.V{}
```
andydotxyz, line 12: As noted above, jsonvalue.V isn't part of the API any more, so probably best not in the docs.

andydotxyz (changes requested):
I think the API docs should be updated to reflect removing jsonvalue.V from the public contract.

andydotxyz:
Just circling back to the `Int` discussion as well. I know that JSON only has "number" but would it not be a common use-case that people want to store/communicate integer values?

data/binding/json.go
Bluebugs: Technically an "Int" in JSON is a float without decimal. I can add the logic for an Int here, but I think we should have more conversion support also in Fyne. If you are ok with that, I will also prepare a PR to add FloatToInt and IntToFloat to fyne data binding.

data/binding/json.go
Bluebugs: Indeed.

Bluebugs:
> Just circling back to the `Int` discussion as well. I know that JSON only has "number" but would it not be a common use-case that people want to store/communicate integer values?

I have added the ability to get an Int data binding for a child property.

andydotxyz (approved):
Nice, thanks for this binding :)

PR #33, opened 2022-03-22
add LimitedWidthEntry widget

kgolding:
An extended Entry widget that sets the width, and limits entry of a given number of chars.

### DIFF
--- a/widget/limited_width_entry.go
+++ b/widget/limited_width_entry.go
@@ -15,7 +15,7 @@ type LimitedWidthEntry struct {
 	CharsWide int
 }
 
-// LimitedWidthEntry returns an extended entry that has a minimum width and is limited
+// NewLimitedWidthEntry returns an extended entry that has a minimum width and is limited
 // to a CharsWide number of characters
 func NewLimitedWidthEntry() *LimitedWidthEntry {
 	entry := &LimitedWidthEntry{}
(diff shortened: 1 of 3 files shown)


andydotxyz:
Thanks for the contribution. Looking at the implementation of `MinSize` it seems like this may be a "FixedWidthEntry" rather than a "LimitedWidthEntry"?

I'd also prefer the use of "runes" instead of "chars" I think, to be consistent with what go does...
Though it does open up the question of how to handle multi-byte characters which are multiple runes but a single codepoint (not fully supported yet, but coming soon). I think that it would be reasonable to expand to handle that later though!

widget/limited_width_entry.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+	return min
+}
+
+// TypedRune is called when this item receives a char event.
+//
+// Implements: fyne.Focusable
+func (e *LimitedWidthEntry) TypedRune(r rune) {
+	if e.CharsWide < 1 || len(e.Text) < e.CharsWide {
```
andydotxyz, line 47: If moving to `rune` rather than `chars` then `len([]rune(e.Text))` will count codepoint/glyph instead of bytes.

andydotxyz (changes requested):
I think this was not marked as changes requested earlier when the changes were requested.

andydotxyz:
Closing this due to inactivity. The same can now be managed with a standard `Entry`, its `Validator` and potentially a replacement MinSize.

PR #35, opened 2022-04-12
Propose graph widget

metal3d:
Propose a `Graph` widget to plot a very simple chart from data.

I will make documentation and screenshot when you tell me that the code is OK.

### DIFF
--- a/cmd/graph_demo/main.go
+++ b/cmd/graph_demo/main.go
@@ -148,7 +148,6 @@ func animateBarChart(chart *charts.BarChart) {
 			chart.SetData(data)
 			chart.Refresh()
 		}
-		// let's play with color
 	}()
 
 }


metal3d:
I'm currently fixing some problems and add more possibilities to play with graphs

e.g. Adding the possibility to draw whatever we want over the graph:
image

andydotxyz:
Oh this is so sweet! Great work @metal3d

metal3d:
I totally changed the hierarchy and how to build "charts". It's now a new package where there is a "LineChart" type. Others charts can be added later (PieChart, HistogramChart...)

Mous events are now managed, we can get a `drawing` overlay to draw on the canvas. Below is an example of drawing lines + circle on the nearest data point based on the mouse position (Graph 4).

image

andydotxyz:
Great work, but I'm not sure about the inclusion of the `Hoverable` and `Tappable` implementations.
As you can see from the implementation it simply implements the interface and passes back to function fields, if they are set. It's basically boiler plate that goes against the Fyne design, which is that to respond to additional events you extend the widget and implement the interface.
Is there a reason you felt it was needed for the core functionality of this widget?

metal3d:
> Great work, but I'm not sure about the inclusion of the `Hoverable` and `Tappable` implementations. As you can see from the implementation it simply implements the interface and passes back to function fields, if they are set. It's basically boiler plate that goes against the Fyne design, which is that to respond to additional events you extend the widget and implement the interface. Is there a reason you felt it was needed for the core functionality of this widget?

I think I'm too used to working with other libraries and that I perceive a kind of "lack" in Fyne in terms of event management. Without this being a harsh criticism on my part, because it may be those other libraries are not optimal in themselves, and their design being their own, it is a point that I have difficulty understanding in terms of facilitation with Fyne. 

I will clarify my thoughts 😉 

GTK, Qt, or even a web page controlled by JavaScript, I can, on every element, add a signal connection (`element.addEventListener("eventname", function)` in JS, `signal_connect("clicked", function)` with GTK, etc...)

In other words, at any time, I can make any element clickable or react to a mouse movement **without having to overload the class**.

I don't mind that we have to do otherwise with Fyne (because there are certainly excellent reasons for it), even if I find the procedure a bit more laborious (you have to implement a whole interface, which doesn't have many methods, but you have to find them and some IDEs have trouble finding them)

As a matter of fact, I have already removed the handlers and modified the examples to use a type overload.

But clearly, as feedback when I was starting with Fyne, I was very surprised that I couldn't "click" on an image - I spent a lot of time at that time realizing that Fyne requires a different development method, and to be honest it was almost so off-putting that I switched to another library. Fortunately, I love Fyne and I'm still here 😝

metal3d:
BarChart and YRange are now OK.

image

andydotxyz:
Very cool indeed, thanks @metal3d. Can you add something to README with an image?
Updating the notes from static-check too would be great, then you'll probably get a more detailed review.

metal3d:
Sorry for the late anwser, post-op recovery is long and I didn't checked all my notifications.
I need to enhance some things and of course I will add examples in README.md :)

andydotxyz:
> Sorry for the late anwser, post-op recovery is long and I didn't checked all my notifications. I need to enhance some things and of course I will add examples in README.md :)

Get well and don't worry we will be here when you're ready to pick it up again.

metal3d:
I guess that you can review again 😄

README.md
```diff
@@ -184,3 +184,49 @@ for validation using an entropy system.
...
+import (
+    "fyne.io/fyne/v2"
+    "fyne.io/x/fyne/widget/charts"
+)
+
+func main() {
+    app := app.New()
+    w := app.NewWindow("Graphs")
+
+    // set data
```
Jacalz, line 222: This code here seems a bit self explanatory and the comment a bit unnecessary, don't you think?

README.md
```diff
@@ -184,3 +184,49 @@ for validation using an entropy system.
...
+    data := []float64{1.2, 5.3, 2.2, 3.4}
+    chart := charts.NewLineChart(nil) // nil = default options
+    chart.SetData(data)
+
+    // add chart object to the view
```
Jacalz, line 227: Same thing here. I think the code is self explanatory and the comment a bit unnecessary.

cmd/graph_demo/main.go
```diff
@@ -0,0 +1,154 @@
...
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+	"fyne.io/x/fyne/widget/charts"
+)
+
+func main() {
+	app := app.New()
+	w := app.NewWindow("Graphs")
+
+	graphWidgets := make([]fyne.CanvasObject, 0)
```
Jacalz, line 19:
```suggestion
	graphWidgets := []fyne.CanvasObject{}
```
Or perhaps this one?
```suggestion
	var graphWidgets []fyne.CanvasObject
```
I don't think you necessarily need to use make to create an empty slice.

cmd/graph_demo/main.go
```diff
@@ -0,0 +1,154 @@
...
+			chart.Options().FillColor = initialColor
+			if mean > 50 {
+				chart.Options().FillColor = theme.ErrorColor()
+			} else if mean > 45 {
+				chart.Options().FillColor = theme.PrimaryColor()
+			}
+			chart.SetData(data)
+			chart.Refresh()
+		}
+		// let's play with color
```
Jacalz, line 151: Did this slip in?

cmd/graph_demo/pointergraph.go
```diff
@@ -0,0 +1,83 @@
...
+type LineChartWithMouse struct {
+	*charts.LineChart
+	mouseOver bool
+}
+
+// NewLineChartWithMouse creates a new line chart with mouse events handlers.
+func NewLineChartWithMouse() *LineChartWithMouse {
+	return &LineChartWithMouse{
+		LineChart: charts.NewLineChart(nil),
+		mouseOver: false,
```
Jacalz, line 24: The default value for a bool is false. You can remove the field.
```suggestion
```

Jacalz (changes requested):
Nice work! Thanks for all of this. I quickly went through the code and found some suggestions and improvements. This does warrant a more thorough review but I will have quite a few busy weeks ahead and figured that this was better than nothing :)

Btw, could you standardise your comments a bit? Some starts with capital letter and some don't. I would prefer if all started with caps.

cmd/graph_demo/main.go
Arsen6331: The second one is probably better. It requires no allocations until the actual `append()` call. There's no reason to use RAM before it's needed, especially when the change is so simple.

Arsen6331: It might also be worth considering using `make()` here and setting the capacity because the amount of elements seems to be known, and seems to always be 6 in this case, so `make([]fyne.CanvasObject, 0, 6)`. That way, it will only require one allocation and no copying. Without this, it is likely that `append()` will end up having to allocate a new slice, and copy all the elements over to the new slice multiple times, and it would also likely end up using more memory if `append()` allocates more than required (which it probably will).

Overall, this isn't really a big deal, but it's nice not to use unneeded resources and time for no reason.

README.md
metal3d: Because it's for the documentation, I often prefer to explain things step by step

README.md
metal3d: Fixed in https://github.com/fyne-io/fyne-x/pull/35/commits/6df8e53c249f231fe899d737617b8d377bb06ad9

README.md
metal3d: Fixed in https://github.com/fyne-io/fyne-x/pull/35/commits/6df8e53c249f231fe899d737617b8d377bb06ad9

cmd/graph_demo/main.go
metal3d: yes, fixed in https://github.com/fyne-io/fyne-x/pull/35/commits/b4dbd38eaed639cb65150f744ab7a2b6969d7d8b

Jacalz:
@metal3d Did you get anywhere with the rewrite? It would be very cool to get this (or an improved version if you are working on it) landed :)

Bluebugs:
Yes, but not completely finished. Can be found at https://github.com/metal3d/fyne-x/tree/propose-graph-widget2 .

PR #36, opened 2022-05-13
Propose responsive layout

metal3d:
The responsive layout proposes a "bootstrap imitation" behavior.

The usage is not documented as I'm afraid that it could be a bit more complex to describe than a widget in the README.

The `cmd/responsiveLayout` demo shows a "good" example.


metal3d:
I'm sure I do something wrong with paddings, let me draft this PR the time I check :)

layout/responsive.go
```diff
@@ -0,0 +1,288 @@
...
+// Responsive register the object with a responsive configuration.
+// The optional ratios must
+// be 0 < ratio <= 1 and  passed in this order:
+//      Responsive(object, smallRatio, mediumRatio, largeRatio, xlargeRatio)
+// They are set to previous value if a value is not passed, or 1.0 if there is no previous value.
+// The returned object is not modified.
+func Responsive(object fyne.CanvasObject, breakpointRatio ...float32) fyne.CanvasObject {
+
+	// Create configuration for this object
+	responsiveConfigurations[object] = NewResponsiveConf(breakpointRatio...)
```
Bluebugs, line 309: You might want to put a mutex lock around the access to this map.

layout/responsive.go
```diff
@@ -0,0 +1,288 @@
...
+//      )
+func NewResponsiveLayout(o ...fyne.CanvasObject) *fyne.Container {
+
+	r := &ResponsiveLayout{
+		configuration: make(map[fyne.CanvasObject]ResponsiveConfiguration),
+		mutex:         &sync.Mutex{},
+	}
+
+	for _, ob := range o {
+		if conf, ok := responsiveConfigurations[ob]; !ok {
```
Bluebugs, line 290: It would be good to hold a lock here to protect the access to responsiveConfigurations.

Bluebugs (commented):
Pretty cool layout. I feel like it should have its place in Fyne itself, what does everyone think?

layout/responsive.go
metal3d: I read about it and it seems that there is no reason to lock on existence check but I can be wrong. Actually there is a lock at the Set function to add an element and another one to delete the element.

metal3d: Anyway I must ensure this. I will add a lock

metal3d: fixed in https://github.com/fyne-io/fyne-x/pull/36/commits/fef4cbaa692e760d40d9c73ced3c6cb019725e7f

layout/responsive.go
metal3d: fixed in https://github.com/fyne-io/fyne-x/pull/36/commits/fef4cbaa692e760d40d9c73ced3c6cb019725e7f

metal3d: fixed in https://github.com/fyne-io/fyne-x/pull/36/commits/01cfc672cf3af05914a4c0bcbfb604bdab72bf51

layout/responsive.go
metal3d: fixed in https://github.com/fyne-io/fyne-x/pull/36/commits/01cfc672cf3af05914a4c0bcbfb604bdab72bf51

Bluebugs:
Thanks. I am sorry, I should have also looked at the API, and I think there is a bit of a need to pick a word for Configuration as you are using Conf, Config and Configuration in the public API. I am bad at naming and don't have a great suggestion here, maybe @andydotxyz would, but I feel that at least there should be only one for the entire API. Maybe Options is the way to go in go?

metal3d:
> Thanks. I am sorry, I should have also looked at the API, and I think there is a bit of a need to pick a word for Configuration as you are using Conf, Config and Configuration in the public API. I am bad at naming and don't have a great suggestion here, maybe @andydotxyz would, but I feel that at least there should be only one for the entire API. Maybe Options is the way to go in go?

I used "Conf" when names are too long. But yes, we can change to "Options".

Bluebugs:
Arg, very sorry about that. I planned to push to my own repository, but somehow it managed to land in yours. If you don't like the change, please just use --force to remove the commit.

I didn't want to ask you to do the change as I was not very confident with my knowledge of Go and Fyne. I have been uncomfortable with the lock and global, so I have tried to remove it by introducing an intermediate "widget" that will just redirect rendering to the original fyne.CanvasObject, but will keep the responsive information for the Layout(). Let me know what you think of this.

And again very sorry about this push.

metal3d:
@Bluebugs no problem, I leaved the option to make maintainers able to push on PR, that's the goal. I prefer having modification and reverting them if problem instead of managing several repository to merge ;)

You fixes are nice. I'm taking a look on them.

Bluebugs (approved):
I think this is in good shape with a small public API and the bar on Fyne-X shouldn't be as high as on Fyne. I think that this should eventually get into Fyne as part of improving cross platform automatic layout, but doing it here give us a space to experiment and get feedback before committing strongly to a new API in Fyne.

Bluebugs:
@Jacalz do you have an idea why those platform_tests status have not been reported?

Jacalz:
> @Jacalz do you have an idea why those platform_tests status have not been reported?

It usually relates to a PR based on a PR that is behind the branch that it wants to merge into (if there have been changes to the Action between those commits). That does however not seem very likely to be the culprit given that the latest CI change was five months ago. Strange.

@andydotxyz would you mind checking that the required checks are the correct ones?

andydotxyz:
> @andydotxyz would you mind checking that the required checks are the correct ones?

Looks like the config was out of date, fixed

PR #40, opened 2022-06-15
Add configuration options to map widget

EricNeid:
Added some properties to configure map widget. This allows to set a custom tile source and added the ability to enable/disable the map control widgets (zoom/move).

### DIFF
--- a/widget/map.go
+++ b/widget/map.go
@@ -29,11 +29,22 @@ type Map struct {
 	zoom, x, y int
 
 	cl *http.Client
+
+	TileSource          string // url to download xyz tiles (example: https://tile.openstreetmap.org/%d/%d/%d.png)
+	EnableOsmDisclaimer bool   // enable osm copyright disclaimer
+	EnableZoomButtons   bool   // enable zoom buttons
+	EnableMoveButtons   bool   // enable move map buttons
 }
 
 // NewMap creates a new instance of the map widget.
 func NewMap() *Map {
-	m := &Map{cl: &http.Client{}}
+	m := &Map{
+		cl:                  &http.Client{},
+		TileSource:          "https://tile.openstreetmap.org/%d/%d/%d.png",
+		EnableOsmDisclaimer: true,
+		EnableZoomButtons:   true,
+		EnableMoveButtons:   true,
+	}
 	m.ExtendBaseWidget(m)
 	return m
 }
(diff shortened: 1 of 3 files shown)


widget/map.go
```diff
@@ -29,11 +29,22 @@ type Map struct {
 	zoom, x, y int
 
 	cl *http.Client
+
+	TileSource          string // url to download xyz tiles (example: https://tile.openstreetmap.org/%d/%d/%d.png)
+	EnableOsmDisclaimer bool   // enable osm copyright disclaimer
+	EnableZoomButtons   bool   // enable zoom buttons
+	EnableMoveButtons   bool   // enable move map buttons
```
Bluebugs, line 36: I think making those field directly accessible and public might be a bit problematic as they could be modified during the life of the widget with unexpected result for the user.

I would think introducing a Config{} or Options{} structure and a new NewMapWithConfig would be a better design pattern. In which case I would make also the http.Client{} part of that structure too.

andydotxyz: This is a cool idea. Another possibility would be to just use setters.

If you go with public fields (even if it's an Options struct) then remember zero values are the normal defaults - so if it should be visible by default then maybe the field is `HideZoomControls`

EricNeid: Sounds good. I like the idea that the configuration and the widget are separated. Now that I think of, maybe the options-pattern could be useful for this use case. 

Something like this:

```go
type MapOption func(*Map)

func WithOsmTiles() MapOption {
	return func(m *Map) {
		m.tileSource = "https://tile.openstreetmap.org/%d/%d/%d.png"
		m.enableOsmDisclaimer = true
	}
}

func NewMapWithOptions(opts ...MapOption) *Map {
	// defaults
	m := &Map{
		cl:                  &http.Client{},
		tileSource:          "https://tile.openstreetmap.org/%d/%d/%d.png",
		enableOsmDisclaimer: true,
		EnableZoomButtons:   true,
		EnableMoveButtons:   true,
	}
	m.ExtendBaseWidget(m)
	for _,opt := range opts {
		opt(m)
	}
	return m
}

m := xwidget.NewMapWithOptions(
	xwidget.WithOsmTiles(),
)
```

EricNeid:
I updated the implementation with new pattern and added some tests for the map widget.

widget/map.go
```diff
@@ -29,15 +29,76 @@ type Map struct {
...
+		m.hideMoveButtons = !enable
+	}
 }
 
 // NewMap creates a new instance of the map widget.
 func NewMap() *Map {
-	m := &Map{cl: &http.Client{}}
+	m := &Map{
+		cl:              &http.Client{},
+		tileSource:      "https://tile.openstreetmap.org/%d/%d/%d.png",
```
andydotxyz, line 82: Can we avoid this duplication of important strings by delegating to the `Options` code you added perhaps?

widget/map.go
```diff
@@ -29,15 +29,76 @@ type Map struct {
 	zoom, x, y int
 
 	cl *http.Client
+
+	tileSource      string // url to download xyz tiles (example: "https://tile.openstreetmap.org/%d/%d/%d.png")
+	hideDisclaimer  bool   // enable copyright disclaimer
+	disclaimerLabel string // label for disclaimer (example: "OpenStreetMap")
```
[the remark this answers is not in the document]
EricNeid: I fixed the naming. 
Yes, the option is included mainly when using non-osm tiles. But it is also useful if you want a custom implementation for the attribution and not depend on the one that is provided with this widget.

widget/map.go
EricNeid: Good idea. I updated NewMap() with your suggestion.

andydotxyz:
Looking good I think, Happy @Bluebugs?
Just need to get the tests passing and we can review it properly.

Bluebugs:
I like it, once the golint warning are addressed I will gladly approve this PR.

andydotxyz (approved):
Great thanks

Bluebugs (approved):

PR #42, opened 2022-07-14
Support custom items in the completion entry

andydotxyz:
Basically exposing the create/update for internal list.
Supports doing just about anything in the list of results when populating.

### DIFF
--- a/widget/completionentry.go
+++ b/widget/completionentry.go
@@ -67,6 +67,9 @@ func (c *CompletionEntry) ShowCompletion() {
 
 	if c.navigableList == nil {
 		c.navigableList = newNavigableList(c.Options, &c.Entry, c.setTextFromMenu, c.HideCompletion)
+	} else {
+		c.navigableList.UnselectAll()
+		c.navigableList.selected = -1
 	}
 	holder := fyne.CurrentApp().Driver().CanvasForObject(c)
 
--- a/widget/completionentry.go
+++ b/widget/completionentry.go
@@ -14,6 +14,9 @@ type CompletionEntry struct {
 	Options       []string
 	pause         bool
 	itemHeight    float32
+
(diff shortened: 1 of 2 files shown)


widget/completionentry.go
```diff
@@ -123,25 +130,38 @@ type navigableList struct {
...
+		customCreate:    create,
+		customUpdate:    update,
 	}
 
 	n.List = widget.List{
 		Length: func() int {
 			return len(n.items)
 		},
 		CreateItem: func() fyne.CanvasObject {
+			if fn := n.customCreate; fn != nil {
```
Bluebugs, line 155: If at creation of n, we defaulted customCreate to create a label. There would be no need at all here for any indirection and we could just set customCreate as the type for CreateItem. Might be good to rethink the name of customCreate here in that case.

widget/completionentry.go
```diff
@@ -123,25 +130,38 @@ type navigableList struct {
...
+				return fn()
+			}
 			return widget.NewLabel("")
 		},
 		UpdateItem: func(i widget.ListItemID, o fyne.CanvasObject) {
+			if fn := n.customUpdate; fn != nil {
```
Bluebugs, line 161: Same apply here,i guess.

Bluebugs (changes requested):

widget/completionentry.go
andydotxyz: If we did as you propose how would someone reset it later if required? Removing this check and user cannot set the value to nil or it would panic.

Bluebugs: I may have missed something, but n is an internal object with customCreate and customUpdate being captured by it with no way to be changed from the outside from what I understand.

Bluebugs (approved):

PR #43, opened 2022-07-17
Calendar widget

itsjustdel:
A calendar widget which returns a [time](https://pkg.go.dev/time) on date selected.
calendarex


vit1251:
Feature request: highlight box/label with specific color of date, start week setup, i18n and l10n support.

Jacalz:
> Feature request: highlight box/label with specific color of date, start week setup, i18n and l10n support.

Let’s stick to something small and easy to review first. Then we can expand as we feel that features are missing. As for i18n and l10n, it isn’t supported upstream in Fyne so adding it for just one widget seems unnecessary.

Jacalz:
I will try to review this in one of the coming days, if I can :)

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+	padding  float32     = 0
+	cellSize float64     = 32
+)
+
+type calendarLayout struct {
+	cols int
+}
+
+func newCalendarLayout(s float64) fyne.Layout {
+	cellSize = s
```
Jacalz, line 31: We should avoid global variables if we can.

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+}
+
+//MinSize sets the minimum size for the calendar
+func (g *calendarLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
+	rows := g.countRows(objects)
+	return fyne.NewSize(float32(cellSize+float64(padding))*7, float32(cellSize+float64(padding))*float32(rows))
+}
+
+type calendar struct {
```
Jacalz, line 95: It would be a lot more useful to have the type exported. It is more in line with other Fyne widgets and is a lot more easy to use for the developer using it.

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+	monthNext     *widget.Button
+	monthLabel    *widget.RichText
+
+	day   int
+	month int
+	year  int
+
+	dates *fyne.Container
+
+	callback func(time.Time)
```
Jacalz, line 110: We usually call these callback functions `OnX` to indicate that it runs on a specific action. How about `OnSelected` in this case?

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+
+	//add spacers if week doesn't start on Monday
+	for i := 0; i < dayIndex-1; i++ {
+		buttons = append(buttons, layout.NewSpacer())
+	}
+
+	for d := start; d.Month() == start.Month(); d = d.AddDate(0, 0, 1) {
+
+		dayNum := d.Day()
+		s := fmt.Sprint(dayNum)
```
Jacalz, line 132: It is a lot faster and more efficient to use the `strconv` functions here.

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+// NewCalendar creates a calendar instance
+func NewCalendar(cT time.Time, callback func(time.Time)) *calendar {
+	c := &calendar{day: cT.Day(), month: int(cT.Month()), year: cT.Year(), calendarTime: cT, callback: callback}
+	c.ExtendBaseWidget(c)
+
+	return c
+}
+
+// NewCalendarAtPos creates a calendar instance and shows at given position
+func NewCalendarAtPos(canvas fyne.Canvas, position fyne.Position, calendarTime time.Time, callback func(time.Time)) *calendar {
```
Jacalz, line 226: This needs to be removed as it isn’t how layouts in Fyne work. It is the parent container and layout that controls where components are located. You should not say to the widgets where it should be drawn.

widget/calendar.go
```diff
@@ -0,0 +1,247 @@
...
+		callback: callback}
+
+	c.ExtendBaseWidget(c)
+	widget.ShowPopUpAtPosition(c, canvas, position)
+
+	return c
+}
+
+// ShowAtPos creates a pop up with instantiated calendar
+func (c *calendar) ShowAtPos(canvas fyne.Canvas, pos fyne.Position) {
```
Jacalz, line 244: This needs to be removed as it isn’t how layouts in Fyne work. It is the parent container and layout that controls where components are located. You should not say to the widgets where it should be drawn.

Jacalz (changes requested):
Thanks for working on this. I have included a few comments to improve the code standard to what we usually go for Fyne components :)

itsjustdel:
Thank you for the review :) I'll start making these changes and update.

widget/calendar.go
matwachich: But what if we want to display it as PopUp? Above an Entry for example.
Could be usefull

Jacalz: The popup function (https://pkg.go.dev/fyne.io/fyne/v2/widget#ShowPopUpAtPosition) just takes a `CanvasObject`. I don’t think you even need to wrap it in a widget.

itsjustdel:
I have made changes according to your lovely suggestions, forgive me if I have gone off track, I'm new to go and fyne and also to contributing to open source projects in general, so keep me right :)

Bluebugs:
Seems a bit of documentation is missing for the Calendar type. Also it would be nice to include the picture of what the widget look like in the README.md file the same way it was done for responsive layout.

README.md
```diff
@@ -57,6 +57,24 @@ Community contributed widgets.
...
+
+To use create a new calendar with a given time and a callback function:
+```go
+calendar := widget.NewCalendar(time.Now(), onSelected, cellSize, padding)
+```
+Once the date is selected you can use the callbackFunction to use the time however you wish:
+```go
+func onSelected(t time.Time) {
+	fmt.Println(t.Format("Mon 02 Jan 2006"))
+}
```
Jacalz, line 75: This documentation seems unintuitive to me. It doesn't really say how the callback is used. How about just showing how a calendar is created. Look at the other widgets in the README and see how they do it.

widget/calendar.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (g *calendarLayout) countRows(objects []fyne.CanvasObject) int {
+	count := 0
+	for _, child := range objects {
+		if child.Visible() {
+			count++
+		}
+	}
+
+	return int(math.Ceil(float64(count) / float64(7)))
```
Jacalz, line 38: You might want to move that 7 out to a global constant like `daysPerWeek` or something along those lines. I know that I told you to avoid global variables but constants are absolutely fine to have global (as they never change their value and improve code readability).

widget/calendar.go
```diff
@@ -0,0 +1,229 @@
...
+			col++
+		}
+		i++
+	}
+}
+
+//MinSize sets the minimum size for the calendar
+func (g *calendarLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
+	rows := g.countRows(objects)
+	return fyne.NewSize((float32(g.cellSize)+g.padding)*7, (float32(g.cellSize)+g.padding)*float32(rows))
```
Jacalz, line 87: The variable `g.cellSize` is already a `float32`. You don't need to convert it.

widget/calendar.go
```diff
@@ -0,0 +1,229 @@
...
+func (c *Calendar) calendarObjects() []fyne.CanvasObject {
+	textSize := float32(8)
+	columnHeadings := []fyne.CanvasObject{}
+	for i := 0; i < 7; i++ {
+		j := i + 1
+		if j == 7 {
+			j = 0
+		}
+
+		var canvasObject fyne.CanvasObject = canvas.NewText(strings.ToUpper(time.Weekday(j).String()[:3]), color.NRGBA{0xFF, 0xFF, 0xFF, 0xBF})
```
Jacalz, line 170: It is better to use the colours from the `theme` package. Also, you can use `:=` here instead of `var`.

widget/calendar.go
```diff
@@ -0,0 +1,229 @@
...
+
+	c.dates = container.New(newCalendarLayout(c.cellSize, c.padding), c.calendarObjects()...)
+
+	dateContainer := container.NewVBox(nav, c.dates)
+
+	return widget.NewSimpleRenderer(dateContainer)
+}
+
+// NewCalendar creates a calendar instance
+func NewCalendar(cT time.Time, onSelected func(time.Time), cellSize float32, padding float32) *Calendar {
```
Jacalz, line 217: It would be better if the `cellSize` just could be computed out of the size of the widget. That's more in line with how we do sizing in Fyne.

I'd also prefer if the padding could use `theme.Padding()` from Fyne instead.

widget/calendar.go
```diff
@@ -0,0 +1,229 @@
...
+// CreateRenderer returns a new WidgetRenderer for this widget.
+// This should not be called by regular code, it is used internally to render a widget.
+func (c *Calendar) CreateRenderer() fyne.WidgetRenderer {
+	c.monthPrevious = widget.NewButtonWithIcon("", theme.NavigateBackIcon(), func() {
+		c.month--
+		if c.month < 1 {
+			c.month = 12
+			c.year--
+		}
+		c.monthLabel.ParseMarkdown(c.monthYear())
```
Jacalz, line 189: Why are you parsing markdown here and everywhere else? It seems to me like it work just fine without it in a regular `widget.Label`.

Jacalz (changes requested):
Thanks for the updates. I only managed a short review today but U hope to have a more thorough look towards the end of the week.

widget/calendar.go
itsjustdel: The problem with using `theme.padding()` is the date buttons are too spaced out if you want any padding at all in the app. In fact, my examples had the calendar padding at 0. I think I will remove the padding for the moment.

widget/calendar.go
itsjustdel: Good question, a hangover from the app this widget was initially designed for, working well with `widget.Label` instead

itsjustdel:
Thank you for taking the time to review my PR, your comments were helpful, however I will need some more time to think about computing the `cellsize` from the size of the widget. 
I simplified the readme and linked to a demo example, any other suggestions for the readme would be welcome, cheers!

Jacalz (changes requested):
Thanks for your fixes. I understand that the cellSize computation can require some thinking. Take your time.

I just has a very quick look through this. As I said earlier, I'll review this more thoroughly later :)

widget/calendar.go
andydotxyz: If we want it to be packed with no padding that is cool - but I think omit the `padding` parameter as @Jacalz is right that's not very Fyne.

itsjustdel:
I have made changes to how the widget is sized and have tried to simplify how the calendar / UI tracks which date it is on.
A starting date can now be given to the calendar too. Looking forward to your feedback

cmd/calendar_demo/main.go
```diff
@@ -20,9 +20,12 @@ func main() {
 	l.Alignment = fyne.TextAlignCenter
 	d := &date{instruction: i, dateChosen: l}
 
-	calendar := xWidget.NewCalendar(time.Now(), d.onSelected, 32)
+	c := container.NewVBox(i, l)
```
andydotxyz, line 23: Why bring this definition forward? It could be created at the end as it was before could it not? Fewer lines of code and avoid the `append`.

itsjustdel: Good point, I don't know I changed that.

cmd/calendar_demo/main.go
```diff
@@ -0,0 +1,42 @@
+package main
+
+import (
+	"time"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/app"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/widget"
+	xWidget "fyne.io/x/fyne/widget"
```
Jacalz, line 10: We have used `xwidget` before. I think it is best to stay consequent.
```go
```

widget/calendar.go
```diff
@@ -0,0 +1,192 @@
...
+	return buttons
+}
+
+func (c *Calendar) dateForButton(dayNum int) time.Time {
+	oldName, off := c.currentTime.Zone()
+	return time.Date(c.currentTime.Year(), c.currentTime.Month(), dayNum, c.currentTime.Hour(), c.currentTime.Minute(), 0, 0, time.FixedZone(oldName, off)).In(c.currentTime.Location())
+}
+
+func (c *Calendar) monthYear() string {
+	return c.currentTime.Month().String() + " " + strconv.Itoa(c.currentTime.Year())
```
Jacalz, line 131: I think you should be able to use https://pkg.go.dev/time#Time.Format here.

Jacalz (changes requested):
This is starting to look very good now. I just have a few small suggestions :)

itsjustdel:
Current appearance:
image

Jacalz (approved):
Thanks for all your hard work on this. I’m happy with it in its current form.
Let us see what @andydotxyz thinks before we merge though :)

itsjustdel:
Very exciting! Thank you for your time on this!

andydotxyz:
Nice job, thanks @itsjustdel :)

andydotxyz (approved):
Very nice. 
Maybe in the future it can get integrated into Fyne and a date picker input that also shows the date spinner on mobile apps :)

PR #48, opened 2022-11-16
Propose desktop integration

metal3d:
This is a first attempt to have a desktop integration theme.

I made more tests on Gnome than KDE, it works at least with Gnome 40 and 42. Best results with Adwaita themes (light or not).

The theme also reacts on settings changes (gsettings, using dbus).

For icons, there are many problems that can happen because rasterX package is not able to parse everything. To fix the issues, I detect the installation of Inkscape or ImageMagick and I temporarily convert problematic icons to PNG. That's very fast.

The example in `cmd/desktop_demo` shows some widget and file view (to see that the file/dir icons are taken from Gnome or KDE icon theme).

It "just work", but I'm of course open for suggestions

PS: I will add the documentation and screenshot in README.md later.


theme/desktop/kde.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// setFont sets the font for the theme.
+func (k *KDETheme) setFont() {
+
+	if k.fontConfig == "" {
+		return
+	}
+	// the fontline is in the form "fontline,size,...", so we can split it
+	fontline := strings.Split(k.fontConfig, ",")
```
Jacalz, line 166: I think you might want to use `strings.SplitN` here given that you only care about the two first ones

theme/desktopEnvironmentLinux.go
```diff
@@ -0,0 +1,35 @@
...
+	wm := os.Getenv("XDG_CURRENT_DESKTOP")
+	if wm == "" {
+		wm = os.Getenv("DESKTOP_SESSION")
+	}
+	wm = strings.ToLower(wm)
+
+	switch wm {
+	case "gnome", "xfce", "unity", "gnome-shell", "gnome-classic", "mate", "gnome-mate":
+		return desktop.NewGnomeTheme(-1, desktop.GnomeFlagAutoReload)
+	case "kde", "kde-plasma", "plasma":
```
Jacalz, line 28: You might want "lxqt" (if that it their exact string in this case) in there as well. I believe it is relatively widely used :)

Jacalz (changes requested):
Thanks for working on this. I just had a very quick look through and noticed a few things that can be improved :)

theme/desktopEnvironmentLinux.go
metal3d: Should be fixed 😄

metal3d:
Just to mention, now that https://github.com/fyne-io/fyne/pull/3414 is in progress, we should not try to detect variant. At least, the way I did it earlier in this proposition.

For KDE, the theme is taken from `kdeglobals`, and I'm not sure that it will be used like this in future releases if they use Adwaita.

For Gnome, the theme is parsed from `gjs` calls, using Adwaita, the colors are OK, but the variant changes needs that the user changed `legacy` theme - that's not OK for me...

IMHO, later, we need to propose an Adwaita theme. That's a pity for Linux users, but the futur Fyne version changed the color scheme.

Later, when Fyne 2.2 will be outdated, we have some new colors to parse (overlay colors, and so on)

andydotxyz:
If you are looking at the possible changes of a Fyne update then hold on - 2.3 is due in a week or so.

metal3d:
@andydotxyz I actually changed this in my local environment, I mean that the theme change detection is now removed, but I just discovered something that will be complex to manage.

Explanation of the problem, and possible resolution:

Gnome changed so many things that this is very complicated to follow. Gnome 43 (and partially Gnome 42) is now using libAdwaita. There is a few changes of color scheme between versions, but that's not the problem.

The issue is that this PR uses `gjs` to read GTK (v3 or v4) theme.  The "gnome Adwaita theme" is not a GTK theme...
AFAIK there is no way to detect Gnome Theme color scheme.

Gnome installs a GTK Adwaita theme to make it possible for legacy GTK applications to follow the Gnome theme. But it is not directly bound to Gnome. The change should be made manually... And, so, my current PR make Fyne applications to be managed as a "legacy GTK application".

I've got a solution that works locally:

- I created an Adwaita theme (with light and dark variants, taken from the documentation)

Then, loading `FromDesktop` theme:

- if the "gnome theme name" is an Adwaita theme, I load the Fyne theme I've made—so it is bound to the color-scheme (with changes detection from Fyne 2.3 version)
- if not, so I continue to use the GTK theme, but it doesn't fit the Gnome theme. It follows the GTK theme (like all other legacy GTK apps actually...)

I also added some "options", to select what we can grab from the theme (icons, font, scaling, all)

This works great, but there are tricky things I've made.

metal3d:
Oh, I think I was wrong in the way I saw it working.
Actually, Adwaita is the main and "non-interchangeable" theme of Gnome, it has two color palettes, and that's it. Gnome applications can then use CSS to adapt their interface.

Older applications, based on GTK 3 or 4, have an associated theme. Either, by default, Adwaita (light), or a theme installed by the user.

So, in fact, my idea is not a bad one.

Only, we have two solutions, and I think that's pretty much how we should look at it:
- I created an Adwaita theme, we can use it directly
- if we ask for `FromDesktopEnvironment`:
    - if Gnome is in version >= 42
       - we activate the Adwaita theme (which reacts very well to the change of color palette)
       - depending on the options (flags), we take the icon theme, the scaling, the font, etc...
    - otherwise:
       - I find the GTK theme activated
       - I do my own thing to adapt to the theme
       - ==> but it will not react to the change of color-scheme, it is the "legacy" mode)

andydotxyz:
> Actually, Adwaita is the main and "non-interchangeable" theme of Gnome, it has two color palettes, and that's it. Gnome applications can then use CSS to adapt their interface.

Yup, that was my understanding too - they ripped out theming and (only very recently) allow theme-like CSS overrides to placate some devs that need more control.

PR #49, opened 2022-12-04
WIP - present a method to use native file selector

metal3d:
Even if the Fyne file dialogs are OK, it nice to be able to use the native file selector as it proposes some native actions as "create a directory", use bookmarks, etc.

Zenity and KDialogs are commonly installed in Linux. This work in progress proposes a way to use native dialogs.

NOTE: this could be extended to info dialogs, alerts, and many others.

### DIFF
--- a/cmd/native_dialog/main.go
+++ b/cmd/native_dialog/main.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"log"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/app"
+	"fyne.io/fyne/v2/widget"
+	"fyne.io/x/fyne/dialog"
+)
+
+func main() {
+	app := app.New()
+	w := app.NewWindow("Hello")
+
+	button := widget.NewButton("Click Me", func() {
+		dialog.NewFileSelector(func(file fyne.URIReadCloser, err error) {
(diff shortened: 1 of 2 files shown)


Bluebugs:
I like this. I actually think that it might be right to have it in the main repository. Their has been demand for native file dialog on Windows and MacOS too. No idea what are others opinion on this, and I would be interested to hear it.

Jacalz:
I came across https://github.com/ncruces/zenity. It might be useful for implementing this more cleanly across multiple platforms.

andydotxyz:
> I came across https://github.com/ncruces/zenity. It might be useful for implementing this more cleanly across multiple platforms.

Are you sure? Does that not rely on an external binary being installed?

Jacalz:
> > I came across https://github.com/ncruces/zenity. It might be useful for implementing this more cleanly across multiple platforms.
> 
> Are you sure? Does that not rely on an external binary being installed?

It does, but so does this PR at the moment. I don't mean that it is a good candidate for being in the main `fyne` repository but rather that it might be a better solution than what this PR currently does.

andydotxyz:
I am leaning back in this direction for file dialogs in the main repo - particularly for Windows and macOS.
Should we pick this up over in `fyne` to avoid duplicating the dialog APIs?

PR #50, opened 2023-01-10
Fix rendering of transparent GIFs

ObaraEmmanuel:
I was using the animated GIF in my projects and noticed it was not correctly rendering GIFs with transparent sections. Further investigation showed that is was because of the use an improper [Porter-Duff operator](http://ssp.impulsetrain.com/porterduff.html#:~:text=In%20the%20Porter%2FDuff%20compositing,the%20image%20is%20partially%20there.). The operator in use before was `Over` which means on subsequent renders, transparent section would contain pixels from the previous frame. This PR is a simple fix that uses the `Src` operator instead. This ensures the new frame overwrites the previous one completely allowing transparent regions to be rendered correctly on subsequent frames.
This change will not result in any regressions because for non-transparent GIFs the two operators (`Over` & `Src`) behave essentially the same way.

### DIFF
--- a/widget/gif.go
+++ b/widget/gif.go
@@ -113,7 +113,7 @@ func (g *AnimatedGif) Start() {
 	g.runLock.Unlock()
 
 	buffer := image.NewNRGBA(g.dst.Image.Bounds())
-	draw.Draw(buffer, g.dst.Image.Bounds(), g.src.Image[0], image.Point{}, draw.Over)
+	draw.Draw(buffer, g.dst.Image.Bounds(), g.src.Image[0], image.Point{}, draw.Src)
 	g.dst.Image = buffer
 	g.dst.Refresh()
 
@@ -132,7 +132,7 @@ func (g *AnimatedGif) Start() {
 				if g.isStopping() {
 					break
 				}
-				draw.Draw(buffer, g.dst.Image.Bounds(), srcImg, image.Point{}, draw.Over)
+				draw.Draw(buffer, g.dst.Image.Bounds(), srcImg, image.Point{}, draw.Src)
 				g.dst.Refresh()
 
 				time.Sleep(time.Millisecond * time.Duration(g.src.Delay[c]) * 10)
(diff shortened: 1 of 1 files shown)


andydotxyz:
Are you certain? It's my understanding that from the git spec a frame can omit pixels which is why we must draw over the previous...

Perhaps there is a difference between a transparent pixel and a mixing pixel?

ObaraEmmanuel:
Here is a section from the w3c GIF spec, See point 2. Maybe there is another spec you were working with?

```
Since a  GIF  data  stream  can  contain  multiple  images,  it  is

   necessary  to  describe  processing and display of such a file.  Because

   the image descriptor allows  for  placement  of  the  image  within  the

   logical  screen,  it is possible to define a sequence of images that may

   each be a partial screen, but in total  fill  the  entire  screen.   The

   guidelines for handling the multiple image situation are:

   1.  There is no pause between images.  Each is processed immediately  as

       seen by the decoder.

   2.  Each image explicitly overwrites any image  already  on  the  screen

       inside  of  its window.  The only screen clears are at the beginning

       and end of the  GIF  image  process.   See  discussion  on  the  GIF

       terminator.
```

ObaraEmmanuel:
> Are you certain? It's my understanding that from the git spec a frame can omit pixels which is why we must draw over the previous...
> 
> Perhaps there is a difference between a transparent pixel and a mixing pixel?

Maybe the spec says to draw over I am not sure. But the GIFs seemed to render properly everywhere else except in fyne

ObaraEmmanuel:
Okay, we need to factor in disposal for each frame for the correct handling. The GIF struct has a disposal array with disposal info for each frame which can be used. Let me apply the correct Porter-Duff operator based on this disposal attribute and see the results

andydotxyz:
That sounds like what I figured - thanks for digging in further

andydotxyz (approved):
That turned out to be quite epic, thanks!

PR #51, opened 2023-01-21
Fix GIF starting and stopping

ObaraEmmanuel:
Currently, once the GIF is stopped it can never be started again. This is because the logic for starting and stopping fails to set the `running` flag to false when stop is called. The `break` is not enough to exit the nested loop. This causes the GIF to remain in the `stopping` state forever after the `Stop()` method is called. This PR exits the nested loop correctly and sets the `running` flag to false allowing the GIF to be started again.


andydotxyz (approved):
Nice catch, thanks!

PR #52, opened 2023-01-22
Massive filetree speedup

andydotxyz:
Cache all the list, canlist and sorting calls for much better performance


andydotxyz:
Review period lapsed with no comment, passing

Issue #38, opened 2022-05-27
CompletionEntry crash on fast input


itsjustdel:
### Describe the bug:

Out of index error on fast input

### To Reproduce:
Steps to reproduce the behaviour:
1. Type four letters e.g "edin" quickly (difficult to get it to crash consistently)
   OR press the "simulate fast typing" button. This gives the same error as typing quickly

2. See error
```
panic: runtime error: index out of range [3] with length 1

goroutine 18 [running, locked to thread]:
fyne.io/x/fyne/widget.newNavigableList.func3(0x7ff7e1196dd8?, {0x7ff7e1196150?, 0xc0000c48c0?})
        C:/Users/Del/go/pkg/mod/fyne.io/x/fyne@v0.0.0-20220524111314-5922941e6ea8/widget/completionentry.go:145 +0xf7
fyne.io/fyne/v2/widget.(*listLayout).setupListItem(0xc0000a81e0, 0xc00038ea10, 0x3)   
        C:/Users/Del/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/widget/list.go:411 +0x9b       
fyne.io/fyne/v2/widget.(*listLayout).updateList(0xc0000a81e0, 0x1)
        C:/Users/Del/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/widget/list.go:456 +0x418      
fyne.io/fyne/v2/widget.(*listLayout).Layout(0x41600000000000f8?, {0x0?, 0xc0004295f8?, 0x7ff7e0ab5fb9?}, {0xecc1d0?, 0xc0?})
        C:/Users/Del/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/widget/list.go:369 +0x1e       
fyne.io/fyne/v2/internal/driver/common.updateLayout({0x7ff7e1195970?, 0xc00019a0c0?})
        C:/Users/Del/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/internal/driver/common/canvas.go:512 +0x70
fyne.io/fyne/v2/internal/driver/common.(*Canvas).EnsureMinSize.func1(0xc001002ac0)    
        C:/Users/Del/go/pkg/mod/fyne.io/fyne/v2@v2.1.0/internal/driver/common/canvas.go:97 +0x235
[...]

PR #53, opened 2023-02-16
Fix crash in CompletionEntry

sharki13:
Fix for https://github.com/fyne-io/fyne-x/issues/38

From time to time there is a crash during typing something to entry, while list of options is modified basis on entered text.
While `UpdateItem` is executed, `n.items` slice is shorter than index `i` indicates, which lead to crash.


widget/completionentry.go
```diff
@@ -162,6 +162,9 @@ func newNavigableList(items []string, entry *widget.Entry, setTextFromMenu func(
 				fn(i, o)
 				return
 			}
+			if i < 0 || i >= widget.ListItemID(len(n.items)) {
```
andydotxyz, line 165: Thanks for this fix, but I do wonder - should we guard the `fn` call above as well? It won't crash immediately but we may be leading the callback to crash or get unexpected callback data...

andydotxyz (commented):
Thanks - just asking what you think about a potential modification.

widget/completionentry.go
andydotxyz: What do you think @sharki13 could you move it so the `fn` callback does not get invalid data as well?

andydotxyz:
Just pinging on this to see if the suggested change could be made so this can be landed?
Will also require rebasing to latest `main` as the versions have updates for checks.

widget/completionentry.go
andydotxyz: Can this questions be answered @sharki13 ? It does seem like a good thing to have fixed.

andydotxyz:
Closing as stale and a new PR exists

PR #55, opened 2023-03-29
DiagramWidget

pbrown12303:
Revised the DiagramWidget and its supporting widgets. Provisions for decorations on links, floating text annotations on links, resizing nodes, and dragging nodes around. Links can connect to nodes and other links (e.g. you can create a link between links) Infrastructure in place for multi-segment links, different decorations on links.


widget/diagramwidget/.gitarchive/COMMIT_EDITMSG
Bluebugs, line 1: I think all the content of .gitarchive should not be here.

widget/diagramwidget/LICENSE
Bluebugs, line 1: This is the same as fyne-x license and shouldn't be necessary.

widget/diagramwidget/README.md
Bluebugs, line 1: I think the readme should be just about the diagram widget.

Bluebugs (dismissed):
This is a massive amount of work. Really cool. It would be nice to have a bit more documentation.

charlesdaniels:
So for some background context, Paul took over (with my consent) a half-completed graph widget from my staging/testing/messing around repo, [fynehax](https://git.sr.ht/~charles/fynehax). Most of the code in 931b1397fc889a356b984c8ec610123fad1e8091 is mine (he and I discussed this directly - the code was committed in this way with my explicit OK), and the subsequent commits are all Paul. 

I do want to give a big shoutout to Paul for polishing this up and adding some really slick new features to make it into something really useful!

cmd/diagramdemo/main.go
```diff
@@ -0,0 +1,137 @@
...
+	"fyne.io/fyne/v2/app"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/widget"
+)
+
+var forceticks int = 0
+
+func forceanim() {
+
+	// XXX: very naughty -- accesses shared memory in potentially unsafe
```
charlesdaniels, line 20: IIRC I did this because I wanted a way to see the layout changing in realtime for debugging purposes. That said, this could probably be made safe if we wrapped forceticks in a mutex.

cmd/tabledemo/main.go
```diff
@@ -0,0 +1,80 @@
+package main
```
charlesdaniels, line 1: This demo as well as the "table" widget should be removed. I think this was pulled in by mistake. I wrote this table before Fyne had a first-party table widget, so it's not needed now methinks.

cmd/viewportdemo/main.go
```diff
@@ -0,0 +1,112 @@
+package main
```
charlesdaniels, line 1: The viewport widget and demo **really need to go**. Aside from being unfinished, it relies on Fyne's renderer to do clipping which means it leaks VRAM very badly. The more you zoom in, the more memory is needed for generating the textures from the scene.

widget/diagramwidget/connectionpad.go
```diff
@@ -0,0 +1,226 @@
...
+	"fyne.io/fyne/v2/widget"
+)
+
+const (
+	pointPadSize float32 = 10
+	padLineWidth float32 = 3
+)
+
+// ConnectionPad is an interface to a connection area on a DiagramElement.
+type ConnectionPad interface {
```
charlesdaniels, line 20: Should include an example of using `ConnectionPad` in the demo, unless it's already in there and I just missed it?

charlesdaniels: nevermind, I see it actually is used :)

widget/diagramwidget/forcelayout.go
```diff
@@ -0,0 +1,84 @@
+package diagramwidget
```
charlesdaniels, line 1: I'm not sure how well this really works. I feel like when I left off it was still pretty rough. Maybe you patched it up a bit?

We may want to consider using `gonum/graph` for layout now or in the future https://pkg.go.dev/gonum.org/v1/gonum@v0.12.0/graph/layout

Perhaps it would make sense to define a `GraphLayouter` interface and refactor the force layout to implement it, that way folks can plug in their own layout engines without having to fork the library?

widget/diagramwidget/geometry/geometry.go
```diff
@@ -0,0 +1,2 @@
+// Package geometry implements various useful geometric primitives.
```
charlesdaniels, line 1: I think there is some overlap between this geometry library and some of the new helpers in Fyne v2. Perhaps we could refactor to use those methods where they are available.

I wrote this geometry stuff back in the Fyne v1 days when the selection of helpers was more limited.

If there are helpers in this library that aren't already upstreamed, perhaps we can look at landing them in Fyne proper.

charlesdaniels:
[Screencast from 03-29-2023 06:39:12 PM.webm](https://user-images.githubusercontent.com/11738895/228683626-80648d2d-e2ae-4ed6-8f1d-8e816b3cce03.webm)

(see attached video)

I see you made it so the nodes are resizable, which is super cool! However when resizing a node, it's possible for it to be "pushed" to the right or down, though not to the left or up. Should detect if the node is already at it's min size and do a NO-OP in this situation to avoid accidentally pushing the nodes around.

widget/diagramwidget/geometry/geometry.go
charlesdaniels: I looked into this a little. It looks like the geometry helpers are a less plentiful than I had remembered in my mind. I only see [Vector2](https://pkg.go.dev/fyne.io/fyne/v2#Vector2).

Considering, I'd propose to keep the geometry stuff as it is in this PR and revisit hoisting these libraries to a top-level package within `fyne-x`, or possibly up-streaming some of them if the core team thinks that's appropriate.

jackclark31:
Not sure if this is the place to leave info.. I have been looking at using Fyne to work out some "GUI" ideas over using React/Web stuff.. more like Admin app stuff. One of the things I was interested in was this NodeRed style graph drag/drop nodes with connections. I saw a few posts about this particular component for Fyne. 

I'll ask here.. but if there is a better place, email forum, reddit forum, etc.. please let me know.

Basically from the video this looks pretty close to what I would like to use. I am wondering if the "work" area that the widgets are placed on.. is itself placeable in any Fyne window? Is it a separate larger widget that handles zoom in/out (or if not.. planned to), scroll so more widgets can be added for bigger widget flows? 

The widget itself.. is it customizable.. e.g. colors (background, text), can the main content area have pictures, text, etc positioned? Like.. the way I envision a widget is it has input ports on the left, and output ports on the right (so to speak). Can you specify the placement of the ports or connection points? As well, can you specify WHAT can be connected to (or from) so that there can be visual. indicators as you drag a connector over a connection point if it can be allowed to be connected? 

Is there a potential top title area above the main box, that allows for drag handle to drag the widget around? 

How customizable is the widget shape? Can it be smaller rectangular.. to allow collapsing them for example?

Lastly, can you put a box around a bunch to group them up.. and then collapse/expand that box to organize the nodes more?

Of all I listed above.. anything that is not possible.. can it be done and/or planned on?

Thank you.

charlesdaniels:
@jackclark31 I'd suggest joining the #fyne channel in the Gophers slack. If you go to fyne.io, there's a link to the channel plus a signup link for the Gophers chat all the way in the bottom on the right.

> Basically from the video this looks pretty close to what I would like to use. I am wondering if the "work" area that the widgets are placed on.. is itself placeable in any Fyne window?

Should be. I didn't test for that specifically when I was writing the initial version. I'm not sure if Paul tested that specifically either. I'd be surprised if it didn't work though as it does implement the `fyne.Widget` interface.

> Is it a separate larger widget that handles zoom in/out (or if not.. planned to)

You might be able to hack it by overriding the scaling level? Normally that's set for an entire app, not sure if it can be overridden on a per-widget basis. You should ask this in the Slack channel. 

> scroll so more widgets can be added for bigger widget flows?

BTW the demonstrated diagram widget already supports panning, so you can have more stuff off the edge of the screen. Fyne itself also supports both horizontally and vertically scroll-able containers.

---

I'm going to leave the rest to @pbrown12303 since he's taking this over. I've been away from Fyne and from this particular project long enough that I don't think I have terribly useful answers. Note that I think Paul is on vacation and unreachable at the moment.

Bluebugs:
I tried to ridiculously put it on the web. If you have wasm, it work: https://diagram.ddlm.me/ (just a bit slow, but still super cool).

andydotxyz:
> [Screencast from 03-29-2023 06:39:12 PM.webm](https://user-images.githubusercontent.com/11738895/228683626-80648d2d-e2ae-4ed6-8f1d-8e816b3cce03.webm)
> 
> 
> 
> (see attached video)
> 
> 
> 
> I see you made it so the nodes are resizable, which is super cool! However when resizing a node, it's possible for it to be "pushed" to the right or down, though not to the left or up. Should detect if the node is already at it's min size and do a NO-OP in this situation to avoid accidentally pushing the nodes around.

Good idea! You made me think and I realised the same bug exists in FyneDesk window resize!

andydotxyz:
> is itself placeable in any Fyne window?

All Fyne widgets (by design) can be placed in any window on any app :).

andydotxyz:
> not sure if it can be overridden on a per-widget basis.

The scale of a window is set for the whole window. Content can be manipulated to scale up using maths in the app - but only canvas primitives, not the standard widgets. You can see how I did that in the Slydes app (presentations scale to fit the space as you'd expect) https://github.com/andydotxyz/slydes

pbrown12303:
@jackclark31 You are describing the exact use case I have in mind. The widget is, indeed, intended to be used as a component of a larger interface. I am using it to implement an IDE for a visual language (a separate project). What I have submitted is a work in progress - I am looking for feedback regarding the style in which it is implemented before I add more functionality.

The diagram since the diagram itself is unbounded in size, I want to add zoom and pan to the DiagramWidget.

I would like to make the widgets (Diagram, Node, and Edge) extensible - I am looking for feedback on the best mechanism for doing  this. I think subclassing is out (only because they already extend BaseWidget). I think I will end up adding interfaces for this purpose (which will, of course, limit the directions in which customization is possible).

You mention one extension I am working on: addint Point connection areas to nodes. The infrastructure is already there. I just need to add the interface to add them to a node. What I am wrestling with is how (what logic to use) to move them when the node scales. Do I just move them proportionally? But what happens when the interior remains fixed in size (i.e. doesn't scale)? Or is this situation not worth considering.

The diagram widget is set up with its own theme. The DiagramElements (nodes and links) take their default values from this theme. Some of these values (e.g. color) can be overridden at the individual diagram element level (as is typical in diagrams)

andydotxyz:
A widget that has extended BaseWidget can still be extended. All the core widgets are set up like this.

jackclark31:
Can we merge this soon.. I'd like to try this out and start getting a handle on how to use it and maybe contribute if possible. Thank you.

jackclark31:
@pbrown12303 I have a whole ton of ideas. :D. Not sure you have ever played with a video editor suite Davinci Resolve (free to download and use too.. which is amazing given how the free version is on par with high end products.. but that's out of scope :D). It has two tabs.. Fusion and Color (pages they are called). Both employ "nodes" and if you got a chance to play with that for a few minutes, you can see how you the nodes are visually very easy and pleasing to work with. They mimic Node Red (or similar). Each type of node has various ins/outs that work with some nodes and dont work with others (e.g. the visual indication of the "port" as you try to connect one node/port to it (so some sort of go code logic callback for each port would be needed to determine if the TYPE (struct, interface?) being dropped is allowed or not.. and more so.. could add functionality to allow the entire node to change color for example.. imagine you start to drag a point from a port of one node.. ALL other nodes in the work area that can NOT accept that disable.. or turn red.. so its easily visual to see that the node you're trying to connect from.. can't connect to those nodes.. and more so.. the nodes that CAN accept it might turn green, and the specific ports might flash that accept it). 

As you have no doubt seen in various other tools like UML tools where they have a box.. not just a little rectangular node red style node, and its got a title area box at the top, perhaps another connected box at the bottom, and little ports along either side. Usually we're look at left to right movement, but there shouldn't be a reason any port on any side couldn't accept connections.

Grouping would be a big thing.. along with expand/collapse of groups. Imagine a tool that has multiple workflows.. that themselves connect at some point. So you can work on a few nodes.. group them (bounding box around them selects them.... right click.. popup with menu option to group, etc). So the ability to be able to expand/collapse groups, then zoom in/out and scroll to move/jump around to another group, zoom in, expand, and work again.. that is a huge time saver for various types of tools. 

So.. me personally.. I love YAML format for describing things like this. You can easily define a node with various aspects like in/out ports, connections (thus dependencies), and so on. BUT.. Go not being dynamic load capable (dang it.. why is that the ONE feature they have in the std lib that is half baked.. either fix it for all platforms or remove it completely).. it would be hard without reflection or something to indicate the function in Go to call when a node is connected, or the type a given port might represent, etc. 

So I am thinking we're kind of stuck with Go code only that represents nodes and functionality behind ports, etc?

But it is imperative that the overall component has some way to export the entire set of nodes (and all connections, etc) so that there is a way to save and load them. That is why I was thinking yaml (or json).

Jacalz:
> Can we merge this soon.. I'd like to try this out and start getting a handle on how to use it and maybe contribute if possible. Thank you.

We can not merge anything in an incomplete state. Before merging, it will have to be reviewed and those comments need to be addressed in the code before approval.

widget/diagramwidget/.gitignore
```diff
@@ -0,0 +1,21 @@
+# Mac OS X files
+.DS_Store
+
+# Binaries for programs and plugins
+*.exe
```
andydotxyz, line 5: these files won't be in the widget package. I suspect that this file could be  removed - as we keep a .gitignore at the root.

widget/diagramwidget/README.md
```diff
@@ -0,0 +1,81 @@
+# Fyne DiagramWidget
```
andydotxyz, line 1: Can you also include something into the top level README so it is more discoverable?

widget/diagramwidget/diagram.go
```diff
@@ -0,0 +1,251 @@
...
+	"image/color"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/driver/desktop"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+var Globaldiagram *DiagramWidget
```
andydotxyz, line 13: I appreciate that this is part of a workaround as described below but we should not have public global APIs.
The workaround should not use globals anyhow as the refresh should apply to the window of the widget - you may have more than one instance in an application.

widget/diagramwidget/diagram.go
```diff
@@ -0,0 +1,251 @@
...
+type linkPinPair struct {
+	link *DiagramLink
+	pin  *LinkPoint
+}
+
+type DiagramWidget struct {
+	widget.BaseWidget
+
+	// ID is expected to be unique across all DiagramWidgets in the application.
+	ID string
```
andydotxyz, line 39: Is this needed? the `DiagramWidget` itself is unique, an additional ID should not be needed.

widget/diagramwidget/diagram.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// Diagrams may want to use a different theme and variant than the application. The default value is the
+	// applicaton's theme and variant
+	DiagramTheme fyne.Theme
+	ThemeVariant fyne.ThemeVariant
+	Offset       fyne.Position
+
+	// DesiredSize specifies the size which the graph widget should take
+	// up, defaults to 800 x 600
+	DesiredSize fyne.Size
```
andydotxyz, line 49: This is an anti-pattern for Fyne widgets.
The desired size is controlled by the application and container it uses for the widget, not from the widget instance itself.

widget/diagramwidget/diagram.go
```diff
@@ -0,0 +1,251 @@
...
+
+	Nodes                          map[string]*DiagramNode
+	Links                          map[string]*DiagramLink
+	selection                      map[string]DiagramElement
+	diagramElementLinkDependencies map[string][]linkPinPair
+
+	// TODO Remove DummyBox when fyne rendering issue is resolved
+	DummyBox *canvas.Rectangle
```
andydotxyz, line 57: Can you please make the workaround a private API so it can be removed without any impact when we resolve the issue?

widget/diagramwidget/diagram.go
```diff
@@ -0,0 +1,251 @@
...
+		ThemeVariant:                   fyne.CurrentApp().Settings().ThemeVariant(),
+		DesiredSize:                    fyne.Size{Width: 800, Height: 600},
+		Offset:                         fyne.Position{X: 0, Y: 0},
+		Nodes:                          map[string]*DiagramNode{},
+		Links:                          map[string]*DiagramLink{},
+		DummyBox:                       canvas.NewRectangle(color.Transparent),
+		selection:                      map[string]DiagramElement{},
+		diagramElementLinkDependencies: map[string][]linkPinPair{},
+	}
+	dw.DummyBox.SetMinSize(fyne.Size{Height: 50, Width: 50})
```
andydotxyz, line 73: No size needed for the dummy I think?

widget/diagramwidget/geometry/r2/box.go
```diff
@@ -0,0 +1,175 @@
...
+//    (1) A.X,A.Y +------+ A.X+S.X,A.Y (2)
+//                |\     |
+//                | \    |
+//                |  \   |
+//                |   \S |
+//                |    \ |
+//                |     \|
+// (3)A.X,A.Y+S.Y +------+ A.X+S.X,A.Y+S.Y (4)
+//
+type Box struct {
```
andydotxyz, line 19: Would it not be worth defining this as a `Box{fyne.Position, fyne.Size}`?
We have the types and maths in the core package

widget/diagramwidget/geometry/r2/line.go
```diff
@@ -0,0 +1,128 @@
...
+// Line describes a line in R2
+//
+//	(1) A.X,A.Y  +
+//	              \
+//	               \
+//	                \
+//	                 \
+//	                  + A.X+S.X,A.Y+S.Y (2)
+type Line struct {
+	// A defines the basis point of the line
```
andydotxyz, line 12: Like box you can use our provided types

widget/diagramwidget/geometry/r2/vec2.go
```diff
@@ -0,0 +1,90 @@
+// Package r2 implements operations relating to objects in R2.
+package r2
+
+import (
+	"math"
+)
+
+// Vec2 implements a vector in R2
+type Vec2 struct {
```
andydotxyz, line 9: It may not be needed to use this, there is a `fyne.Vector` created since this code started.

andydotxyz (changes requested):
It is great to see this coming together.
I have provided some comments inline regarding "idiomatic Fyne" aspects.

I see also that you have combined `fyne.Theme` aspects but also `colour.Color` directly - you might want to consider using the theme more and `fyne.ThemeColorName` so it can be extended in a more Fyne way. Just a thought though, not as important as the other points.

Overall I think this adds more public APIs than required - especially the introduction of a new `geometry` and `r2` packages opens up a whole new API surface that `fyne-x` may not wish to commit to supporting in the long term.

cmd/viewportdemo/main.go
pbrown12303: Removed

widget/diagramwidget/diagram.go
pbrown12303: Made local

widget/diagramwidget/README.md
pbrown12303: Moved all of the DiagramWidget documentation into the top-level Readme

widget/diagramwidget/diagram.go
pbrown12303: ID is needed here in the same way that ID is needed for tree widget entries: to facilitate a) retrieving widgets and b) mapping widgets to the underlying information they represent (which is often more complicated than a simple value
binding).

widget/diagramwidget/diagram.go
pbrown12303: Done

widget/diagramwidget/diagram.go
pbrown12303: I don't know. Do canvas objects with zero size actually get refreshed?

.gitignore
```diff
@@ -40,3 +40,4 @@ Session.vim
 tags
 # Persistent undo
 [._]*.un~
+cmd/diagramdemo/__debug_bin.exe
```
Jacalz, line 43: This is less specific and catches more cases:
```suggestion
*.exe
```

go.mod
```diff
@@ -1,14 +1,44 @@
 module fyne.io/x/fyne
 
-go 1.14
+go 1.20
```
Jacalz, line 3: Can you not keep the Go module mostly as it was? We should not have 1.20 as the minimum version. Just include any new dependencies you need for this change so we can keep the changes minimal.

widget/diagramwidget/.gitignore
Jacalz, line 1: You should remove this whole file. Use the `.gitignore` at the root instead.

widget/diagramwidget/geometry/r2/box.go
pbrown12303: The r2 package here is mostly about geometry operations. The algorithms in the package (e.g. what is the intersection between the box and a line) use these geometries. Not sure I want to re-write those algorithms at this time. On the other hand, I think it is appropriate to keep all of the r2 stuff local to the widget so that it can be refined/replaced later. I'll work on that.

Jacalz (changes requested):
I added a few notes inline.

widget/diagramwidget/geometry/r2/line.go
pbrown12303: See above comment

cmd/diagramdemo/main.go
pbrown12303: Charles: can you be more specific about your thoughts for the Mutex - what is being locked?

widget/diagramwidget/forcelayout.go
pbrown12303: I am not making any provisions for automatic layout. Diagram editors typically have multiple layout algorithms, but they are explicitly invoked by the user. Anyone building a diagram editor can add their favorite layout algorithms.

widget/diagramwidget/diagram.go
pbrown12303: There is no longer a need for the dummy - it has been removed.

widget/diagramwidget/geometry/r2/box.go
pbrown12303: I don't see a practical way to limit the visibility of the r2 package components. I think we should leave it for now

widget/diagramwidget/geometry/r2/line.go
pbrown12303: Again, the purpose of the r2 package is to perform operations on geometric shapes. Short of revising the fyne geometry package (which is beyond the scope of the DiagramWidget), I don't see another alternative. If the fyne geometry package is updated (a separate project), I would be more than happy to update the DiagramWidget to refer to it rather than the r2 package.

.gitignore
pbrown12303: Done

pbrown12303:
This appears to be waiting for three reviews. Can we please move this forward?

Jacalz:
We are very close to the release of Fyne v2.4.0 and have thus been focusing our time there.  I've also been quite busy recently on top of that so this has had to be pushed back. Will try to look at it beginning of next week.

widget/diagramwidget/connectionpad.go
```diff
@@ -0,0 +1,348 @@
...
+
+// MouseDown responds to mouse down events
+func (rp *RectanglePad) MouseDown(event *desktop.MouseEvent) {
+	connectionTransaction := rp.padOwner.GetDiagram().ConnectionTransaction
+	if connectionTransaction != nil {
+		link := connectionTransaction.Link
+		if link.isConnectionAllowed(connectionTransaction.LinkPoint, rp) {
+			padOwnerPosition := rp.padOwner.Position()
+			pseudoEvent := &fyne.DragEvent{
+				PointEvent: fyne.PointEvent{},
```
Jacalz, line 71: Default value. No need to set it here.

widget/diagramwidget/geometry/r2/vec2.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Length return the vector length
+func (v Vec2) Length() float64 {
+	return math.Sqrt(math.Pow(v.X, 2) + math.Pow(v.Y, 2))
+}
+
+// Dot returns the dot product of vector v and u
+func (v Vec2) Dot(u Vec2) float64 {
+	return v.X*u.X + v.Y + u.Y
```
Jacalz, line 34: Should this not be `v.X*u.X + v.Y*u.Y`?

Jacalz (changes requested):
Well done. This is really cool. I just have a few inline notes but otherwise this seems like a great start :)

cmd/diagramdemo/main.go
```diff
@@ -0,0 +1,187 @@
...
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/widget"
+)
+
+var forceticks int = 0
+
+func forceanim(diagramWidget *diagramwidget.DiagramWidget) {
+
+	// XXX: very naughty -- accesses shared memory in potentially unsafe
+	// ways, this almost certainly has race conditions... don't do this!
```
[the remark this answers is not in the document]
pbrown12303: Removed

README.md
```diff
@@ -80,20 +91,91 @@ calendar := widget.NewCalendar(time.Now(), onSelected, cellSize, padding)
...
+or resized. 
+
+* [demo](../../cmd/diagramdemo/main.go)
+
+<p align="center" markdown="1" style="max-width: 100%">
+  <img src="img/DiagramWidget.png" width="1024" height="880" alt="Diagram Widget" style="max-width: 100%" />
+</p>
 
-### FileTree
+**DiagramElement Interface**
```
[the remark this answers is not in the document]
pbrown12303: Done

README.md
```diff
@@ -80,20 +91,91 @@ calendar := widget.NewCalendar(time.Now(), onSelected, cellSize, padding)
...
-</p>
+This collection should be considered a work in progress. When changes are made,
+serious consideration will be given to backward compatibility, but compatibility
+is not guaranteed. 
 
-```go
-gif, err := NewAnimatedGif(storage.NewFileURI("./testdata/gif/earth.gif"))
-gif.Start()
-```
+The DiagramWidget itself is intended to be incorporated into a Fyne application. It provides a
```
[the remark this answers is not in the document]
pbrown12303: Done

widget/diagramwidget/anchoredtext.go
```diff
@@ -0,0 +1,161 @@
...
+		offset:            r2.MakeVec2(0, 0),
+		ForegroundColor:   theme.ForegroundColor(),
+		referencePosition: fyne.Position{X: 0, Y: 0},
+	}
+	at.displayedTextBinding = binding.NewString()
+	at.displayedTextBinding.Set(text)
+	at.textEntry = widget.NewEntryWithData(at.displayedTextBinding)
+	at.textEntry.Wrapping = fyne.TextWrapOff
+	at.textEntry.Validator = nil
+	at.textEntry.Enable()
```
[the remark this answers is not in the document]
pbrown12303: Removed

widget/diagramwidget/connectionpad.go
pbrown12303: I'm not sure what you are referencing here. Please clarify.

Jacalz: You can remove the line just over my comment. You are setting a field to its default value.

widget/diagramwidget/geometry/r2/vec2.go
pbrown12303: Good catch! I inherited this code and never even read it! I fixed it, but it turns out it's never called. We should revisit the entire geometry thing at some point.

Jacalz: Good plan but let's keep the code correct and not not broken now that we know that it is wrong.

Also FYI, we go by the rule that the reviewer closes their own comments when they deem them as resolved, not the one resolving them.

widget/diagramwidget/connectionpad.go
pbrown12303: Got it. Done.

pbrown12303:
Got it. I need to learn the system!

-- PCB

_______________________________________________________________

Paul C. Brown

Paul C. Brown Consulting LLC - *Specializing in IT Architecture and
Strategy*

"Total architecture is not a choice - it is a concession to reality."


Email: pbrow ***@***.***>***@***.***         Mobile: 518-424-5360



Architecture Books:

-- Succeeding With SOA: Realizing Business Value Through Total Architecture
<http://www.informit.com/authors/bio.aspx?a=53cdb713-432d-4e63-a5b1-3a9ee6fedbeb>

-- Implementing SOA: Total Architecture In Practice
<http://www.informit.com/authors/bio.aspx?a=53cdb713-432d-4e63-a5b1-3a9ee6fedbeb>

-- TIBCO Architecture Fundamentals
<http://www.informit.com/authors/bio.aspx?a=53cdb713-432d-4e63-a5b1-3a9ee6fedbeb>

-- Architecting Composite Applications and Services with TIBCO
<http://www.informit.com/authors/bio.aspx?a=53cdb713-432d-4e63-a5b1-3a9ee6fedbeb>

-- Architecting Complex-Event Processing Solutions with TIBCO
<http://www.informit.com/authors/bio.aspx?a=53cdb713-432d-4e63-a5b1-3a9ee6fedbeb>



The SOA Manifesto: soa-manifesto.or <http://soa-manifesto.org/>g

______________________________________________________________

Jacalz (dismissed):
Looks good to me. Well done.

charlesdaniels:
Wow, this is really awesome! Glad to see this coming together. A serious ~~graph~~ diagram widget is something I've really wanted to see in Fyne ever since I made the first hacky toy in the Fyne v1 days. I'm really tickled that Paul was able to turn it into something real. 

I tried messing with the demo a bit, and it seems to be working quite well. I do have two small nits I want to bring up...

1. It seems like the arrowheads "flicker" when they overlap with the labels. I wonder if perhaps they are fighting over position in the Z-order? I don't remember exactly how that works in the Fyne internals anymore.
2. It seems that the labels can be "dragged" away from the nodes/edges, but continue to be displaced along with the node/edge they were previously "attached" to. Maybe I'm misunderstanding the metaphor, but that seems unintuitive to me and I wanted to be sure it's the intended behavior.

I have attached a video demonstrating both of these behaviors.

[Screencast from 08-15-2023 08:06:14 PM.webm](https://github.com/fyne-io/fyne-x/assets/11738895/47091066-d8a7-49e7-b984-5cc91335c11f)

pbrown12303:
Folks - I have a long-term consulting engagement starting in a couple of weeks. Once that starts, I will have limited bandwidth to address review feedback about the DiagramWidget. Please, if possible, try to complete your reviews while I still have the time to address your change requests.

andydotxyz:
As @charlesdaniels mentioned this really is incredible work, thanks so much for sharing!

Having tried the latest version I have the following thoughts:
*) the flickering of labels as mentioned above seems to happen on every refresh, which should be looked at - however it also shows that refresh is happening every time the mouse moves over the diagram so it will likely be burning a lot of CPU unnecessarily. Can this be avoided?
*) The "drop" point seems to suffer a similar flicker - the area I want to move a link to won't always be highlighted and sometimes it is really difficult to move the mouse one just enough to make it activate before it goes away again.
*) Scrolling does not seem to work quite right - you can drag things out of view but the scroll only activates for the "min size" of the diagram irrespective of where it is placed. When resizing the window it seems to try and pull the diagram back to the visible area but doesn't quite get it right. I would consider aiming for a version without scrolling enabled and it can then be added in a follow up where the details can be resolved to scroll across all content (or beyond?).

I don't know what you feel about this. I would think it could be in good shape if the flicker was resolved and scrolling removed/disabled (temporarily). Great work!

pbrown12303:
I believe I have addressed all outstanding change requests. Flickering has been fixed; Scrolling is now fully implemented; Lines are easier to select.

andydotxyz (changes requested):
Looking amazing, I just added one comment about line arrow angles not updating then I think it could merge from what I see.

andydotxyz:
Oops, looks like a staticcheck issue exists as well.

Jacalz:
Can you please remove the changes to go.mod and go.sum? That seems entirely irrelevant to the changes being discussed. If we are to update anything we need to test all of the repository first

Jacalz:
I did a test update to v2.4.0 recently and noticed that the map widget wasn't working. Let's let that change stew before we update

widget/diagramwidget/diagram.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // Verify that interfaces are fully implemented
-var _ fyne.Tappable = (*drawingArea)(nil)
+var _ fyne.Tappable = (*DrawingArea)(nil)
```
andydotxyz, line 16: Why did this become a public type, and also a public field on the diagram?

pbrown12303:
- Backed out fyne 2.4.0 in go.mod and go.sum. I had just assumed that since 2.4.0 had been released that you would want fyne-x to be built against it, but I understand why you want to keep that separate.
- DrawingArea is public to support application-level testing (e.g. drag-drop) that requires sending simulated Drag, Mouse, and Tap events to the drawing. I added comments to indicate  this intent in the code.
- I fixed the arrowhead rotation bug that @andydotxyz identified. It was only showing up with the fyne 2.4.0 dependency, but now it won't happen at all even after the upgrade.
- AnchoredText is supposed to render with scrolling disabled. It works OK with fyne 2.3.5, but not with 2.4.0. The fix itself has a dependency on fyne 2.4.0, so I put the fix in and commented it out with a note indicating that it needs to be uncommented for 2.4.0. The DiagramWidget will still work with 2.4.0 with the line still commented, but you'll get scrolling on the anchored text labels on the  links.

pbrown12303:
Missed the static test fix for node.go - now updated.

andydotxyz:
> DrawingArea is public to support application-level testing (e.g. drag-drop) that requires sending simulated Drag, Mouse, and Tap events to the drawing. I added comments to indicate this intent in the code.

We should not have public APIs purely for testing. If you want to test items that are not public (totally reasonable) then put the test in the same package as the actual code. In the toolkit main we call these `*_internal_test.go`.

andydotxyz:
Looks like there is a failure but it can be ignored - the uploader is not working with Go 1.14 - but in a separate item we should update this repo to Go 1.17 and Fyne 2.4.0

go.work.sum
Jacalz, line 1: Please delete this file.

Jacalz (changes requested):

go.work.sum
pbrown12303: Removed

widget/diagramwidget/diagram.go
pbrown12303: Application-level testing needs to send simulated Drag, Mouse, and Tap events to the DrawingArea. I put comments in the code indicating that this is the only reason that this class and its access are public.

pbrown12303:
@andydotxyz R.e. the DrawingArea being public: the testing I am referring to is NOT in fyne-x: it is in the application that is using fyne-x. If I am dragging and dropping from a tree node to the drawing area and I want to test that, then the application needs to be able to simulate the related events. Similarly with selecting a toolbar button (to create something) and then tapping in the drawing area to place the created item. The alternative would be to add some "DiagramWidget.Simulate<xxx>Event() methods that would pass the events on to the (now private) drawing area. What do you think?

pbrown12303:
@andydotxyz I'm going to go with the simulated events and make DrawingArea private again. Safer.

widget/diagramwidget/diagram.go
andydotxyz: Why do you need to expose an API for that? See how, for example, the `test` package can `Tap` anything tappable without needing to know its API...

andydotxyz:
> @andydotxyz I'm going to go with the simulated events and make DrawingArea private again. Safer.

You don't need them either. And either way that's a lot of new APIs just for the sake of testing.
We have a `test` package that aims to help with testing - the widgets don't need to code it all up themselves...

For example a random object that is tappable:

```go
obj := myType.content
test.Tap(obj.(fyne.Tappable))
```

No public API or simulation required. If the test package does not help with the event then we could either improve that or just call the handlers directly...

```go
ev := fyne.DragEvent{...}
obj.Drag(ev)
```

for example.

pbrown12303:
@andydotxyz Thank you! I think you have provided the answer I have been seeking. Let me try it tomorrow.

pbrown12303:
Getting close. I see two options:

1) Leave both the drawingArea type and the drawingArea attribute of DiagramWidget private. This requires the following testing approach:

```go
func getTappableDrawingArea(dw *diagramwidget.DiagramWidget) fyne.Tappable {
	return test.WidgetRenderer(dw).Objects()[0].(*container.Scroll).Content.(fyne.Tappable)
}

getTappableDrawingArea(diagram).Tapped(newPointEventAt(100, 100))
```

2) Make DrawingArea a public class but leave the drawingArea attribute of the DiagramWidget private. This allows;
```go
func getDrawingArea(dw *diagramwidget.DiagramWidget) *diagramwidget.DiagramArea {
	return test.WidgetRenderer(dw).Objects()[0].(*container.Scroll).Content.(fyne.Tappable)
}

getDrawingArea(diagram).Tapped(newPointEventAt(100, 100))
```
Which do you think is preferable?

pbrown12303:
I think I have addressed all outstanding change requests.

andydotxyz:
> Getting close. I see two options:
> 
> 1. Leave both the drawingArea type and the drawingArea attribute of DiagramWidget private. This requires the following testing approach:
> 
> ```go
> func getTappableDrawingArea(dw *diagramwidget.DiagramWidget) fyne.Tappable {
> 	return test.WidgetRenderer(dw).Objects()[0].(*container.Scroll).Content.(fyne.Tappable)
> }
> 
> getTappableDrawingArea(diagram).Tapped(newPointEventAt(100, 100))
> ```
> 
>  2. Make DrawingArea a public class but leave the drawingArea attribute of the DiagramWidget private. This allows;
> 
> ```go
> func getDrawingArea(dw *diagramwidget.DiagramWidget) *diagramwidget.DiagramArea {
> 	return test.WidgetRenderer(dw).Objects()[0].(*container.Scroll).Content.(fyne.Tappable)
> }
> 
> getDrawingArea(diagram).Tapped(newPointEventAt(100, 100))
> ```
> 
> Which do you think is preferable?

2 is certainly not something I would go for as it requires public API to test. However maybe you are missing Option 3:

```go
func getDrawingArea(...) fyne.CanvasObject {
    ...
}

getDrawingArea(diagram).(fyne.Tappable).Tapped(...)
```

pbrown12303:
Your option 3 is the appropriate one in this case: the functionality being tested is not the core functionality of the Diagram Widget - it is the functionality of the callbacks that the application has added to the widgets which are going to be in the application's package, not fyne-x. Thanks for all the help here.

Is there anything else I need to do?

andydotxyz:
> Your option 3 is the appropriate one in this case: the functionality being tested is not the core functionality of the Diagram Widget - it is the functionality of the callbacks that the application has added to the widgets which are going to be in the application's package, not fyne-x. Thanks for all the help here.
> 
> 
> 
> Is there anything else I need to do?

Happy to help.

The widget looks really great. With the info above can a few more tests be added? Happy otherwise!

pbrown12303:
I wasn't planning on adding any more tests. The ones that we were discussing were for testing my application-supplied callback functions, not functionality within the Diagram Widget itself. The test events are simply the triggers for the callback functions.

andydotxyz:
Thanks so much, with the tests all passing I'm happy. I think @Jacalz has an outstanding change request - what do you think?

pbrown12303:
Pretty sure I resolved his concern. I had upgraded the go.mod to reference fyne 2.4 and he wanted me to roll that back. I did.

andydotxyz:
> Pretty sure I resolved his concern. I had upgraded the go.mod to reference fyne 2.4 and he wanted me to roll that back. I did.

I agree - but his outstanding change request is blocking the merge.

andydotxyz (approved):

README.md
```diff
@@ -13,13 +13,13 @@ This repository holds community extensions for the [Fyne](https://fyne.io) toolk
 
 This is in early development and more information will appear soon.
 
-## Layouts
```
andydotxyz, line 16: I was about to try and merge this in, but realised this change to README.
A markdown document should have only 1 heading, so the changes here to re-level each item isn't right.

andydotxyz (changes requested):

Jacalz (approved):
I agree with the comment from Andy but otherwise approved. Well done :)

andydotxyz (approved):
Amazing work thanks :)

widget/diagramwidget/anchoredtext.go
```diff
@@ -42,7 +43,7 @@ func NewAnchoredText(text string) *AnchoredText {
 	at.displayedTextBinding.AddListener(at)
 	at.textEntry.Wrapping = fyne.TextWrapOff
 	// TODO After upgrade to fyne 2.4.0, uncomment the following line and add container as an imported package
```
andydotxyz, line 45: I think this TODO should have been removed

widget/diagramwidget/README.md
```diff
@@ -0,0 +1,148 @@
+<p align="center">
```
andydotxyz, line 1: I think this block of HTML is not needed for a sub-readme

andydotxyz (changes requested):
A couple of minor changes in this latest update.
Plus could you please rename the Screenshot file?

pbrown12303:
Requested changes made

widget/diagramwidget/anchoredtext.go
pbrown12303: Done

andydotxyz (approved):
Excellent work, thanks so much for pulling this epic widget together!

PR #56, opened 2023-04-02
Add GridWrap widget

dweymouth:
This is a port from fyne to fyne-x of the GridWrapList component I made for [Supersonic](https://github.com/dweymouth/supersonic) used for the album grid view. I stripped out the use of private APIs (`internal/widget/scroll`, `BaseWidget.super()`) and copied others (`syncPool`).

### DIFF
--- a/widget/gridwrap.go
+++ b/widget/gridwrap.go
@@ -14,20 +14,20 @@ import (
 )
 
 // Declare conformity with Widget interface.
-var _ fyne.Widget = (*GridWrapList)(nil)
+var _ fyne.Widget = (*GridWrap)(nil)
 
-// GridWrapListItemID is the ID of an individual item in the GridWrapList widget.
-type GridWrapListItemID int
+// GridWrapItemID is the ID of an individual item in the GridWrap widget.
+type GridWrapItemID int
 
-// GridWrapList is a widget with an API very similar to widget.List,
-// that lays out items in a wrapping grid similar to container.NewGridWrap.
+// GridWrap is a widget with an API very similar to widget.List,
+// that lays out items in a scrollable wrapping grid similar to container.NewGridWrap.
 // It caches and reuses widgets for performance.
-type GridWrapList struct {
+type GridWrap struct {
 	widget.BaseWidget
 
-	Length     func() int                                          `json:"-"`
-	CreateItem func() fyne.CanvasObject                            `json:"-"`
-	UpdateItem func(id GridWrapListItemID, item fyne.CanvasObject) `json:"-"`
+	Length     func() int                                      `json:"-"`
+	CreateItem func() fyne.CanvasObject                        `json:"-"`
+	UpdateItem func(id GridWrapItemID, item fyne.CanvasObject) `json:"-"`
 
 	scroller      *container.Scroll
 	itemMin       fyne.Size
@@ -35,20 +35,20 @@ type GridWrapList struct {
 	offsetUpdated func(fyne.Position)
 }
 
-// NewGridWrapList creates and returns a GridWrapList widget for displaying items in
+// NewGridWrap creates and returns a GridWrap widget for displaying items in
 // a wrapping grid layout with scrolling and caching for performance.
-func NewGridWrapList(length func() int, createItem func() fyne.CanvasObject, updateItem func(GridWrapListItemID, fyne.CanvasObject)) *GridWrapList {
-	gwList := &GridWrapList{BaseWidget: widget.BaseWidget{}, Length: length, CreateItem: createItem, UpdateItem: updateItem}
+func NewGridWrap(length func() int, createItem func() fyne.CanvasObject, updateItem func(GridWrapItemID, fyne.CanvasObject)) *GridWrap {
+	gwList := &GridWrap{BaseWidget: widget.BaseWidget{}, Length: length, CreateItem: createItem, UpdateItem: updateItem}
 	gwList.ExtendBaseWidget(gwList)
 	return gwList
 }
 
-// NewGridWrapListWithData creates a new GridWrapList widget that will display the contents of the provided data.
-func NewGridWrapListWithData(data binding.DataList, createItem func() fyne.CanvasObject, updateItem func(binding.DataItem, fyne.CanvasObject)) *GridWrapList {
-	gwList := NewGridWrapList(
+// NewGridWrapWithData creates a new GridWrap widget that will display the contents of the provided data.
+func NewGridWrapWithData(data binding.DataList, createItem func() fyne.CanvasObject, updateItem func(binding.DataItem, fyne.CanvasObject)) *GridWrap {
+	gwList := NewGridWrap(
 		data.Length,
 		createItem,
-		func(i GridWrapListItemID, o fyne.CanvasObject) {
+		func(i GridWrapItemID, o fyne.CanvasObject) {
 			item, err := data.GetItem(int(i))
 			if err != nil {
 				fyne.LogError(fmt.Sprintf("Error getting data item %d", i), err)
@@ -62,7 +62,7 @@ func NewGridWrapListWithData(data binding.DataList, createItem func() fyne.Canva
 }
 
 // CreateRenderer is a private method to Fyne which links this widget to its renderer.
-func (l *GridWrapList) CreateRenderer() fyne.WidgetRenderer {
+func (l *GridWrap) CreateRenderer() fyne.WidgetRenderer {
 	l.ExtendBaseWidget(l)
 
 	if f := l.CreateItem; f != nil {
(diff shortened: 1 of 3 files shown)


widget/gridwraplist.go
```diff
@@ -0,0 +1,355 @@
...
+
+	return l.BaseWidget.MinSize()
+}
+
+func (l *GridWrapList) scrollTo(id GridWrapListItemID) {
+	if l.scroller == nil {
+		return
+	}
+	row := math.Floor(float64(id) / float64(l.getColCount()))
+	y := (float32(row) * l.itemMin.Height) + (float32(row) * theme.Padding())
```
Jacalz, line 94: Not a big deal but removing the parentheses makes it slightly easier to read in my opinion (it is of course logically the same per the operation order rules).
```suggestion
	y := float32(row) * l.itemMin.Height + float32(row) * theme.Padding()
```

widget/gridwraplist.go
```diff
@@ -0,0 +1,355 @@
...
+// ScrollToOffset scrolls the list to the given offset position
+func (l *GridWrapList) ScrollToOffset(offset float32) {
+	l.scroller.Offset.Y = offset
+	l.offsetUpdated(l.scroller.Offset)
+}
+
+// GetScrollOffset returns the current scroll offset position
+func (l *GridWrapList) GetScrollOffset() float32 {
+	return l.offsetY
+}
```
Jacalz, line 151: This seems a bit like internal details? What is the use case for them?

widget/gridwraplist.go
```diff
@@ -0,0 +1,355 @@
...
+func (l *gridWrapListRenderer) Destroy() {
+}
+
+func (l *gridWrapListRenderer) Objects() []fyne.CanvasObject {
+	return l.objects
+}
+
+func (r *gridWrapListRenderer) SetObjects(objects []fyne.CanvasObject) {
+	r.objects = objects
+}
```
Jacalz, line 197: I think method is unused?

widget/gridwraplist.go
```diff
@@ -0,0 +1,355 @@
...
+			l.itemMin = f().MinSize()
+		}
+	}
+	layout := &fyne.Container{}
+	l.scroller = container.NewVScroll(layout)
+	layout.Layout = newGridWrapListLayout(l)
+	layout.Resize(layout.MinSize())
+	objects := []fyne.CanvasObject{l.scroller}
+	lr := newGridWrapListRenderer(objects, l, l.scroller, layout)
+	return lr
```
Jacalz, line 79: I think this is a bit more readable.
```suggestion
	layout := &fyne.Container{Layout: newGridWrapListLayout(l)}
	l.scroller = container.NewVScroll(layout)
	layout.Resize(layout.MinSize())
	
	return newGridWrapListRenderer([]fyne.CanvasObject{l.scroller}, l, l.scroller, layout)
```

Jacalz (changes requested):
Well done. This is looking great. I just have a few comments inline about making the code more readable in some places.

We talked a bit on Slack about changing the name to something like `GridWrap`. I think the `List` part makes it a bit unclear.

widget/gridwraplist.go
dweymouth: I use it in my app to be able to reconstruct a new GridWrap(List) with the same data *and* at the same scroll position when the user navigates back to a previous page that had a grid view. In my history state I store the data model only and destroy the widgets. (And in the future to reduce allocations I'll reuse the widgets across different pages, just updating the backing data model, instead of destroying+creating new ones)

widget/gridwraplist.go
dweymouth: Yep! And this could be improved in the Fyne `widget.List` code too since this was copied and adapted from there :)

Jacalz: Ah, I see. PRs are always welcome ;)

widget/gridwraplist.go
Jacalz: That makes sense. I am fine with keeping it there in that case

widget/gridwrap.go
```diff
@@ -0,0 +1,349 @@
...
+	scroller      *container.Scroll
+	itemMin       fyne.Size
+	offsetY       float32
+	offsetUpdated func(fyne.Position)
+}
+
+// NewGridWrap creates and returns a GridWrap widget for displaying items in
+// a wrapping grid layout with scrolling and caching for performance.
+func NewGridWrap(length func() int, createItem func() fyne.CanvasObject, updateItem func(GridWrapItemID, fyne.CanvasObject)) *GridWrap {
+	gwList := &GridWrap{BaseWidget: widget.BaseWidget{}, Length: length, CreateItem: createItem, UpdateItem: updateItem}
```
Jacalz, line 41: There is no need to pass `widget.BaseWidget{}` here.
```suggestion
	gwList := &GridWrap{Length: length, CreateItem: createItem, UpdateItem: updateItem}
```

Jacalz (changes requested):
Well done. I mostly just had one small comment in line. Would you mind adding a few simple tests? With that done, I think this should be ready to land.

dweymouth:
> Well done. I mostly just had one small comment in line. Would you mind adding a few simple tests? With that done, I think this should be ready to land.

Done!

Jacalz:
I tried this using the following code but I am not seeing what I would expect. Am I doing anything wrong?
```go
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/theme"

	"fyne.io/fyne/v2/widget"
)

func main() {
	w := app.New().NewWindow("GridWrap Collection")

	items := 1000

	data := make([]fyne.Resource, items)
	for i := 0; i < items; i++ {
		switch i % 10 {
		case 0:
			data[i] = theme.AccountIcon()
		case 1:
			data[i] = theme.CancelIcon()
		case 2:
			data[i] = theme.CheckButtonIcon()
		case 3:
			data[i] = theme.FileApplicationIcon()
		case 4:
			data[i] = theme.FileVideoIcon()
		case 5:
			data[i] = theme.DocumentIcon()
		case 6:
			data[i] = theme.MediaPlayIcon()
		case 7:
			data[i] = theme.MediaRecordIcon()
		case 8:
			data[i] = theme.FolderIcon()
		case 9:
			data[i] = theme.FolderOpenIcon()
		}
	}

	list := xwidget.NewGridWrap(
		func() int {
			return len(data)
		},
		func() fyne.CanvasObject {
			return widget.NewIcon(theme.WarningIcon())
		},
		func(id xwidget.GridWrapItemID, item fyne.CanvasObject) {
			item.(*widget.Icon).Resource = data[id]
		},
	)
	w.SetContent(list)

	w.Resize(fyne.NewSize(256, 256))
	w.ShowAndRun()
}

```
image

Jacalz:
Ignore my previous comment. I had forgotten a refresh call :/

widget/gridwrap_test.go
```diff
@@ -0,0 +1,103 @@
...
+	list := NewGridWrap(
+		func() int {
+			return len(data)
+		},
+		func() fyne.CanvasObject {
+			icon := widget.NewIcon(theme.DocumentIcon())
+			return icon
+		},
+		func(id GridWrapItemID, item fyne.CanvasObject) {
+			item.(*widget.Icon).Resource = data[id]
```
Jacalz, line 98: The issue that I got was because this code (that I copied) isn't calling refresh. While it isn't technically needed with these tests, adding the refresh will make it easier to add more/rendering tests in the future (and avoids people using the code as an example). 
```suggestion
			item.(*widget.Icon).SetResource(data[id])
```

Jacalz (changes requested):
I just have one minor comment inline. This is looking very good :)

Jacalz (approved):
Thanks. I'm very happy with this. Well done!

Perhaps @andydotxyz or @Bluebugs can review before merging?

andydotxyz (changes requested):
This looks really great, thanks.
However I'm not sure about the data binding aspect.
When you attach a `DataList` it is a list of binding items, it will be triggered if the list changes in dimension, but any child item can be updated when that data changes.
In this implementation those child changes will not be reflected in the UI as the item is not bound, nor is it refreshed on a child change.
I'd recommend removing the binding code for now, landing this, and then working on that separately...

andydotxyz (approved):
Apologies, it seems I may have not read the data item code correctly. Looks good.
Happy to proceed if @Jacalz agrees the data binding code is cool to land together with the main API.

Jacalz:
I think it is fine. It looks to be the same as what we already have in the List widget.

Jacalz:
Thanks for your work on this @dweymouth. We appreciate it. This widget is going to be very useful :)

PR #57, opened 2023-04-18
Adding struct form

aodhan-domhnaill:
Example use case,

```
type Transaction struct {
	gorm.Model
	Timestamp            time.Time `fyne:"field"`
	Amount               float64   `fyne:"field"`
	Description          string    `fyne:"field"`
}

type FinancialData struct {
	db *gorm.DB
}

func (fd *FinancialData) AddTransaction(tr *Transaction) error {
	if err := fd.db.Create(&tr).Error; err != nil {
		return err
	}
	return nil
}

func main() {
	// Open a database connection using the sqlite driver
	db, err := gorm.Open(sqlite.Open("accounts.db"), &gorm.Config{})
	if err != nil {
		panic("Failed to connect to database!")
	}

	fd := data.NewFinancialData(db)

	....

	transactionModal = widget.NewStructForm(
		myWindow.Canvas(),
		data.Transaction{},
		func(s interface{}) {
			fd.AddTransaction(s.(*data.Transaction))
		},
	)

	....
}
```

### DIFF
--- a/widget/struct_form.go
+++ b/widget/struct_form.go
@@ -1,0 +1,146 @@
+package widget
+
+import (
+	"reflect"
+	"regexp"
+	"strings"
+	"time"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/data/binding"
+	"fyne.io/fyne/v2/widget"
+)
+
+type StructForm struct {
+	widget.BaseWidget
+	structType reflect.Type
+	modal      *widget.PopUp
+	canvas     fyne.Canvas
+	fields     []reflect.StructField
+	widgets    []*widget.FormItem
+	bindings   []binding.DataItem
+	submit     func(s interface{})
+}
+
+func NewStructForm(c fyne.Canvas, s interface{}, submit func(s interface{})) *StructForm {
+	sf := &StructForm{}
+	sf.ExtendBaseWidget(sf)
+
+	sf.canvas = c
+	sf.structType = reflect.TypeOf(s)
+	sf.fields = parseFields(sf.structType)
+	sf.widgets = createWidgets(sf.fields)
+	sf.submit = submit
+
+	form := &widget.Form{
+		OnCancel: func() {
+			sf.modal.Hide()
+		},
+		OnSubmit: func() {
+			sf.Submit()
+			sf.modal.Hide()
+		},
+		Items: sf.widgets,
+	}
+	sf.modal = widget.NewModalPopUp(
+		container.NewVBox(form),
+		sf.canvas,
+	)
+
+	return sf
+}
+
+func buildValue(t reflect.Type, fields []reflect.StructField, bindings []binding.DataItem) reflect.Value {
+	val := reflect.New(t)
+
+	for i, f := range fields {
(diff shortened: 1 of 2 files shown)


Bluebugs:
This is pretty cool. Just an idea here, what do you think of changing `field` to actually be the widget type to use. For example, you could have:
```
type Transaction struct {
	Timestamp            time.Time `fyne:"entry"`
	Amount               float64   `fyne:"slider:0-100"`
	Description          string    `fyne:"entry"`
        Paid                 bool      `fyne:"checkbox"`
}
```

I will totally accept the PR with just handling the `entry` type.

widget/struct_form.go
```diff
@@ -0,0 +1,146 @@
...
+	"strings"
+	"time"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/data/binding"
+	"fyne.io/fyne/v2/widget"
+)
+
+type StructForm struct {
```
Bluebugs, line 15: Can you add some documentation for all exported type?

widget/struct_form.go
```diff
@@ -0,0 +1,146 @@
...
+	widget.BaseWidget
+	structType reflect.Type
+	modal      *widget.PopUp
+	canvas     fyne.Canvas
```
Bluebugs, line 19: You do not need to store this value if you only access it from within NewStructForm.

widget/struct_form.go
```diff
@@ -0,0 +1,146 @@
...
+	fields     []reflect.StructField
+	widgets    []*widget.FormItem
+	bindings   []binding.DataItem
+	submit     func(s interface{})
+}
+
+func NewStructForm(c fyne.Canvas, s interface{}, submit func(s interface{})) *StructForm {
```
Bluebugs, line 26: It could be useful to split the function in two, one to create just the form and another to pack it in a dialog.

aodhan-domhnaill:
That's a good idea. I'll work on a change.

andydotxyz:
Isn't this what `dialog.NewForm` is for?

Bluebugs (changes requested):
Can you update the top README file and add an example in cmd?

Bluebugs:
> Isn't this what `dialog.NewForm` is for?

This generate the item of NewForm by using reflection. It could use NewForm internally I guess.

andydotxyz (commented):
This is looking good, but note you have outstanding comments from @Bluebugs.

Just thinking out loud here, but did you consider extending Form instead of BaseWidget? Perhaps the NeeStructForm could even extend form - I'm not sure but I wonder if that is cleaner? (I think the CreateRenderer could directly return Form.CreateRenderer).

andydotxyz:
Just trying to catch up on things, where did this get to @Bluebugs @aodhan-domhnaill ? It's such a cool feature, I am embarrassed that we let it sit.

aodhan-domhnaill:
> Just trying to catch up on things, where did this get to @Bluebugs @aodhan-domhnaill ? It's such a cool feature, I am embarrassed that we let it sit.

Sorry. I kind of forgot about this because life stuff. I can try to pick it up again soon

PR #58, opened 2023-05-17
bluetooth drivers

MatejMagat305:
this is not final version, but skeleton (base, it contains bugs and unrefactor things), I would like for some advise and notice <br>
Goal is bring bluetooth on GO (Fyne) to android


andydotxyz:
This is a really epic piece of work, thanks @MatejMagat305. I love the idea and would be happy to help push it forward. I think the main thing we need to do is figure out how to avoid the API being Android specific.
The API should be agnostic to the OS so we can add more implementations in later.
The two main things this means:
1) remove `android` from the naming
2) remove all the `env` and related contexts from the API (we can use the recent `RunNative` system so just need the `fyne.App` context).

I know this is a big ask, but it will make the result so much more useful over time.
This may also mean that we need more interfaces where the current API is android specific structs, though in some cases it may be possible to switch out the details using compile flags.

Excellent work, thanks!

andydotxyz:
Fyne-x repo will always depend upon Fyne. If you want to offer the opportunity for people to use the Bluetooth work without that you might have to consider putting the contribution outside the fyne-x repository :(.

drivers/bluetooth/functionJVM.go
```diff
@@ -0,0 +1,19 @@
...
+package bluetooth
+
+import (
+	"errors"
+	"fyne.io/fyne/v2/driver"
+)
+
+var runOnJVM = originForm
+
+func originForm(fn func(vm, env, ctx uintptr) error) error {
```
MatejMagat305, line 12: do you agree with thar?

andydotxyz: I don't understand the question. (Or maybe I just don't understand the name "originForm").
Using build tags we can switch in different "runNative" functions so the type assertion in your code is safe.

MatejMagat305: simply I do not consider clean everywhere write cast construction ..., so question is whether I must rewrite it to all function to this form or I can keep this solution (maybe "originForm" rename to "unwrap" or "uncast" ... or the most simply  use `//go:link fyne.io/fyne/internal/.../mobileinit/RunOnJVM` - next question is which one you what you prefer)

andydotxyz: I don't understand what then indirection is trying to achieve? Just call the method "runNative" or similar, and then have a no-op version in a non-android file. Global variables and go:link just make this more complex font they?

MatejMagat305: ok, go:linkname I will not use ...

andydotxyz:
You should not have to use linkname - you can just use RunOnJVM directly, or alias the function name to a local name if that is required for some reason.
It's great to have another platform on here to show it can work - would it be possible to add a third "other" implementation that is a no-op so that apps won't fail to compile when the platform is not supported? (runtime error on init instead of a compile error).

MatejMagat305:
> You should not have to use linkname - you can just use RunOnJVM directly, or alias the function name to a local name if that is required for some reason. It's great to have another platform on here to show it can work - would it be possible to add a third "other" implementation that is a no-op so that apps won't fail to compile when the platform is not supported? (runtime error on init instead of a compile error).

arch sorry, linkname was my experiment ..., yes other will be done, but please I have school exams period and I want make this for windows too, but it will hardest part of this

andydotxyz:
> think should I try something with cgo with help tinygo.org/x/bluetooth or would I try for something like [go-ole/go-ole#198](https://github.com/go-ole/go-ole/issues/198) which seems like pure go implementation?

I guess it's up to you, but I don't think we should be depending upon tiny at the repository level.

Also, if you are adding files from other repositories please do write a comment in the header or commit so it is clear - and double check the license of the source as well.

MatejMagat305:
> I found https://github.com/OpenBluetoothToolbox/SimpleBLE can I ask you, cmd fyne can embed `.jar` or `.dex` file to apk? This jni is too complicated ..., but i do not want to bring something breaking new ... I am asking, because I found also https://git.wow.st/gmp/ble (which put idea in my head) and [golang/sys@66a0560](https://github.com/golang/sys/commit/66a0560e4e097a54e439cdc529e28fcd0f9014e8) (which add interface for windows bluetooth)
> 
> i maybe should reopen, but i do not know ...

I do not know whether it will work as multiplatform, but it is progress ..., but if fyne can embede `.dex` file please write ......

andydotxyz:
Our packaging does not support adding .jar or .dex file into the app at this time

> can I ask you, cmd fyne can embed `.jar` or `.dex` file to apk? This jni is too complicated ..., but i do not want to bring something breaking new ...

No, Fyne package command cannot embed other libraries at this time

PR #59, opened 2023-05-17
ResponsiveCanvasObject interface + object hiding

aodhan-domhnaill:
Two changes,

- Interface for ResponsiveCanvasObject so that custom types can be used
- Resizing behavior is handled by the interface method
- Add hiding behavior so object can be hidden if the window shrinks

First two are self explanatory. The hiding behavior is something that I personally wanted so I can have a "summary/detail" view in a table widget I will am working on.

Let me know if you think I should move the hiding behavior out into another interface.

### DIFF
--- a/layout/responsive.go
+++ b/layout/responsive.go
@@ -61,6 +61,9 @@ const (
 // Breakpoint is a uint16 that should be set from const SMALL, MEDIUM, LARGE and XLARGE.
 type responsiveConfig map[responsiveBreakpoint]float32
 
+// hidableConfig = true if the object should be hidden at that breakpoint
+type hidableConfig map[responsiveBreakpoint]bool
+
 // newResponsiveConf return a new responsive configuration.
 // The optional ratios must
 // be 0 < ratio <= 1 and  passed in this order:
@@ -130,19 +133,18 @@ func (resp *ResponsiveLayout) Layout(objects []fyne.CanvasObject, containerSize 
 
 	// For each object, place it at the right position (pos) and resize it.
 	for _, o := range objects {
-		if o == nil || !o.Visible() {
-			continue
-		}
-
 		// get tht configuration
 		ro, ok := o.(ResponsiveCanvasObject)
 		if !ok {
 			log.Fatal("A non responsive object has been packed inside a ResponsibleLayout. This is impossible.")
 		}
 
-		line = append(line, o) // add the container to the line
-
 		size := ro.ComputeWindowResize(windowSize, containerSize)
+		if !ro.Visible() {
+			continue
+		}
+
+		line = append(line, o) // add the container to the line
 
 		// place and resize the element
 		o.Resize(size)
@@ -255,6 +257,7 @@ type responsiveWidget struct {
 
 	render           fyne.CanvasObject
 	responsiveConfig responsiveConfig
+	hidableConfig    *hidableConfig
 }
 
 var _ fyne.Widget = (*responsiveWidget)(nil)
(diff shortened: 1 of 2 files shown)


andydotxyz:
I don't think the concept of "ResponsiveCanvasObject" works - only widgets can be extended.
The primitive canvas objects and the container type do not support extending. The code will compile but the objects will not appear (because the graphics driver does not know how to draw them).

aodhan-domhnaill:
@andydotxyz I don't know the details, but it works :)

```
func (rt *ResponsiveTable) AddRow(ele ...fyne.CanvasObject) {
	l := layout.NewResponsiveLayout()
	for i, e := range ele {
		row := layout.Responsive(
			e,
			rt.rowConstants[i]*rt.sizeTotals[0],
			rt.rowConstants[i]*rt.sizeTotals[1],
			rt.rowConstants[i]*rt.sizeTotals[2],
			rt.rowConstants[i]*rt.sizeTotals[3],
		)
		if i > 1 {
			row.Hidable(true)
		}
		l.Add(row)
	}
	rt.tableContainer.Add(l)
}
```

The behavior is what I would expect too. The details in my table row hide when I shrink the window

aodhan-domhnaill:
By the way, I'm building widgets and layouts with the goal to get to https://react-bootstrap-table.github.io/react-bootstrap-table2/. I've been breaking them into separate PRs as I slowly build them out.

If you have a different approach to managing PRs, let me know.

andydotxyz:
> @andydotxyz I don't know the details, but it works :)
> 
> 
> 
> ```
> 
> func (rt *ResponsiveTable) AddRow(ele ...fyne.CanvasObject) {
> 
> 	l := layout.NewResponsiveLayout()
> 
> 	for i, e := range ele {
> 
> 		row := layout.Responsive(
> 
> 			e,
> 
> 			rt.rowConstants[i]*rt.sizeTotals[0],
> 
> 			rt.rowConstants[i]*rt.sizeTotals[1],
> 
> 			rt.rowConstants[i]*rt.sizeTotals[2],
> 
> 			rt.rowConstants[i]*rt.sizeTotals[3],
> 
> 		)
> 
> 		if i > 1 {
> 
> 			row.Hidable(true)
> 
> 		}
> 
> 		l.Add(row)
> 
> 	}
> 
> 	rt.tableContainer.Add(l)
> 
> }
> 
> ```
> 
> 
> 
> The behavior is what I would expect too. The details in my table row hide when I shrink the window

I'm not sure that the code sample shows usage of the `ResponsiveCanvasObject` that I was unsure of?

aodhan-domhnaill:
Changed it to extending Widget. Maybe that's better

andydotxyz:
Thanks.

What I am curious about is whether this has to be tied to window size? It breaks the container model we use everywhere else.
Can't it just use the container size and conform with the widget/layout contract without the need for connecting to the current window and assuming it has maths based on the whole space?

aodhan-domhnaill:
https://github.com/fyne-io/fyne-x/blob/master/layout/responsive.go#L112

It's in the existing code. I didn't like it either, but didn't want to break your existing functionality.

andydotxyz (changes requested):
I'm not sure why this change introduces so many new API. Also it seems to move the container layout logic into widgets which isn't how containers generally work.

Perhaps @metal3d can comment on the general idea of hiding at thresholds and how it fits into responsive overall, but this implementation feels like it may not be the best way?

andydotxyz:
Can we close this in favour of #79 my @metal3d ?

aodhan-domhnaill:
Closing in favor of #79

PR #60, opened 2023-05-27
Rewrite of CompletionEntry

matwachich:
simplified code
support variable height items
first item automatically selected
items are selected on hover and remplaced on Tap or Enter
Tests pass

Added completion_demo to cmd

Changed fyne version to latest (2.3.4) in go.mod


widget/completionentry.go
```diff
@@ -2,21 +2,26 @@ package widget
 
 import (
 	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/driver/desktop"
```
matwachich, line 5: Seems like this will be a problem on mobile builds...

andydotxyz: You link link to the abstraction... just don't assume desktop types if it's mobile runtime.

go.mod
```diff
@@ -1,14 +1,14 @@
-module fyne.io/x/fyne
+module github.com/matwachich/fyne-x
```
andydotxyz, line 1: You'll need to revert this - not sure how it slipped in

widget/completionentry.go
```diff
@@ -2,21 +2,26 @@ package widget
...
 	widget.Entry
-	popupMenu     *widget.PopUp
-	navigableList *navigableList
-	Options       []string
-	pause         bool
-	itemHeight    float32
+
+	Options           []string
+	OnCompleted       func(string) string // Called when you select an item from the list ; return different string to override the completion
+	SubmitOnCompleted bool                // True will call Entry.OnSubmited when you select a completion item
```
andydotxyz, line 16: Are you sure this needs to be optional? The `OnSubmitted` call is already optional, so this is kinda doubled up and could just always on can't it?

widget/completionentry.go
```diff
@@ -63,168 +69,202 @@ func (c *CompletionEntry) ShowCompletion() {
...
+
+func (item *completionEntryListItem) CreateRenderer() fyne.WidgetRenderer {
+	return widget.NewSimpleRenderer(item.co)
+}
+
+func (item *completionEntryListItem) Tapped(_ *fyne.PointEvent) {
+	item.parent.setTextFromMenu(item.parent.Options[item.id])
+}
+
+func (item *completionEntryListItem) MouseIn(_ *desktop.MouseEvent)    { item.parent.list.Select(item.id) }
```
andydotxyz, line 268: I think it's not OK that this only happens on desktop instance.
I can't see how mobile apps will be able to select an item.

andydotxyz (changes requested):
Thanks for working on this. Not sure about a few things inline

widget/completionentry.go
matwachich: This will just select an item to show it selected on hover, item completion is not triggered by list.Select, but by item Tapped

widget/completionentry.go
matwachich: OnComplete is important to eventually modifiy the string to be inserted in the Entry.

Imagine what is stored in Options is not exactly what should be inserted in Entry...(*)

SubmitOnComplete is for when we want OnSubmitted to be called uppon completion. We could manually call OnSubmitted inside OnComplete but imagine we are in the case above (*), when in OnComplete the Entry content is not yet right.

widget/completionentry.go
matwachich: You are right, I removed this, and the auto-item-select on hover.
It doesn't really make sens, it was just a personal flavor

geodimm:
Hi, any updates on this? 

I have my own fix for the popupMenu (re)sizing as it's currently broken. This PR also seems to fix at least the size calculation issue so would be nice if it gets merged.

matwachich:
I'm not really working on this PR anymore, but for my project I use matwachich/fynex-widgets AutoComplete entry widget and it works like a charm!

geodimm:
Thanks for the quick reply. I have followed up with https://github.com/fyne-io/fyne-x/pull/95

widget/completionentry.go
andydotxyz: Sorry for letting this drop - I guess I didn't understand the use-case. Both the list and the Entry are user facing, so why should what is input not match what is selected?

p.s. "Completed" is misleading here - It means the action is finished but when you select an item you are not necessarily "done" - Other widgets use different verbs to indicate that a menu item is selected / actioned.

andydotxyz:
Sorry that nobody got to reviewing this before.
Given the #95 and the fork you are working on @matwachich should this be re-based after landing or would you prefer it be closed? The conflicts mean it needs work to continue one way or another...

PR #61, opened 2023-05-27
Minor improvements to widget.GridWrap

Jacalz:
We now make sure that this matches https://github.com/fyne-io/fyne/blob/6558290e7ea3cc2604a5537b08c824593282fe24/widget/list.go#L20 in widget.List. It makes it possible to use int directly and even share the function between gridwrap and list if necessary.

On top of that I just put in some very minor changes to the code.


andydotxyz (approved):
Nice

PR #62, opened 2023-06-09
feat: add pagination widget

lazywhite:


lazywhite:
this pagination widget referenced some web ui component like this https://element.eleme.io/#/en-US/component/pagination, 
pagination usally works with Tables, user can do database query or http request inside OnChange() to get a range of data then update the table.

andydotxyz:
> this pagination widget referenced some web ui component like this https://element.eleme.io/#/en-US/component/pagination

It seems a lot more verbose than what is on this link. And they basically don't have page size as a mandatory part of the UI. A more concise UX would be so much slicker.

lazywhite:
> 

what's your thought about page size?  should I delete page size related code and set it a default value?

lazywhite:
> This looks like quite an interesting widget - but it seems quite text heavy. And the user-editable page size in the main widget seems strange, it could be added to the demo to show what devs can control - but it doesn't seem like an important aspect for users.
> 
> Additionally could the demo command show it in usage not just how the info bar looks? For example how does it actually control what pages are visible etc?

I'm sorry, I forget about this, do you think change page size Entry to  Select is a better idea?

andydotxyz:
> I'm sorry, I forget about this, do you think change page size Entry to Select is a better idea?

I'm not sure that how page size is represented is a big deal - but I do think it should not be a mandatory part of the UI (see my other message) as most instances don't want that to be user customisable.

It is possible that a select will be more controllable than an entry too, but then how would you decide what appropriate values are?

lazywhite:
user can pass a []string as select options;

I think maybe this widget is not suitable for community wide usage,  PR closed

andydotxyz:
A []string of options would be a good plan. Perhaps if this was an empty list you could hide that element?

I'm not sure why you think this isn't a useful widget, folk ask for pagination fairly regularly.

PR #64, opened 2023-06-18
Add a portion layout with percentage pased sizes

Jacalz:
This is a WIP layout to allow a space to be divided up into portions based on percentages in a slice. The example below contains four buttons that are set to take up 30%, 20%, 30% and 10% respectively (with remaining 10% empty at the end).

[Screencast from 2023-06-18 13-56-34.webm](https://github.com/fyne-io/fyne-x/assets/25466657/c2f3f912-091d-46f2-85c4-e5f1915c1869)

### DIFF
--- a/layout/portion.go
+++ b/layout/portion.go
@@ -37,18 +37,30 @@ func (p *HPortion) MinSize(objects []fyne.CanvasObject) fyne.Size {
 		return fyne.NewSize(0, 0)
 	}
 
-	size := fyne.Size{}
-	for _, item := range objects {
-		min := item.MinSize()
-		size.Width += min.Width
-		size.Height = fyne.Max(size.Height, min.Height)
+	if len(objects) == 0 {
+		return fyne.NewSize(0, 0)
 	}
 
-	return size
+	maxMinWidth := float32(0)
+	maxIndex := -1
+	height := float32(0)
+
+	for i := 0; i < len(objects); i++ {
+		min := objects[i].MinSize()
+		height = fyne.Max(height, min.Height)
+
+		if min.Width > maxMinWidth {
+			maxMinWidth = min.Width
+			maxIndex = i
+		}
+	}
+
+	return fyne.NewSize(maxMinWidth/p.Portions[maxIndex], height)
 }
 
 // NewHPortion creates a layout that partitions objects verticaly taking up
 // as large of a portion of the space as defined by the given slice.
+// The portions should be between 0 and 1 but not equal to.
 // The length of the Portions slice needs to be equal to the amount of objects.
 func NewHPortion(Portions []float32) *HPortion {
 	return &HPortion{Portions: Portions}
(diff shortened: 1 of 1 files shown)


Jacalz:
Here's a demo of the `HPortion` layout being used for the code input and receive button in Rymdport:
image

Jacalz:
I believe that this is ready for a review now. Some feedback would be nice and then I plan to add some tests after that.

metal3d:
Isn't it a more or less what is proposed in Responsive layout ?

Jacalz:
I don’t think so. There seems to be a little bit of overlap but this isn’t necessarily much more responsive than any other layouts already present in Fyne. I think both layouts have their place. This one is simpler and useful for specific use cases but the other one is a lot more responsible and can do much more. What do you think?

metal3d:
That's right. Responsive hasn't got exactly the same behavior:

```go
	w.SetContent(
		layout.NewResponsiveLayout(
			layout.Responsive(long, 0.3),
			layout.Responsive(shorter, 0.2),
			layout.Responsive(long2, 0.3),
			layout.Responsive(btn, 0.1),
		),
	)
```

This will scale the widgets more or less the same, but they are not resized vertically.

Note: I also remark that `layour.NewResponsiveLayout` returns a `fyne.Container` - I maybe made an error here...

So, to conclude, I finally like your layout a lot 😄

dweymouth:
I wonder if there's a clean (API-wise) way to support fixed width segments as well, with the percentages/portions applying to the remaining width after accounting for the fixed-width segments. That would really make this layout cover a lot of use cases but even as-is it's a nice one

Jacalz:
> Note: I also remark that layour.NewResponsiveLayout returns a fyne.Container - I maybe made an error here...

You could consider renaming it to container. Breaking changes are allowed in fyne-x ;)

> So, to conclude, I finally like your layout a lot 😄

Cool. Thanks :)


> I wonder if there's a clean (API-wise) way to support fixed width segments as well, with the percentages/portions applying to the remaining width after accounting for the fixed-width segments. That would really make this layout cover a lot of use cases but even as-is it's a nice one

Hmm. Yeah, it seems like that could be hard. It bring up a lot of questions regarding what the fixed width should be defined against (how that scales depending on window sizes etc.) and also how the API itself should look of course. I wonder if perhaps this layout should stay simple and that there is a use case for a specific layout with fixed sizes?

Jacalz:
How should we handle portions, in the slice, that are out of range?
We can either check if the value is `== 0` or `> 1` and handle that in some way. I don't know what's best? Just ignoring? Should zero be allowed?

Also, do we want to handle portions that don't add up to 1?

dweymouth:
> Hmm. Yeah, it seems like that could be hard. It bring up a lot of questions regarding what the fixed width should be defined against (how that scales depending on window sizes etc.) and also how the API itself should look of course. I wonder if perhaps this layout should stay simple and that there is a use case for a specific layout with fixed sizes?

Yeah I guess it'd be trying to do too much with one layout maybe. I was just mentioning this since it's very similar to my ColumnsLayout from Supersonic used in tracklists, which lays out a number of columns where they can be either fixed-width, or flexible width and the flexible width columns split the remaining space equally (like PortionLayout but with equal percentage)

> How should we handle portions, in the slice, that are out of range?
> Also, do we want to handle portions that don't add up to 1?

One way for both of these, as long as portions are positive values, is to normalize them by their sum, so that portions 2.5, 1.5, 1 would be the same as portions 25, 15, 10, which is the same as 50%, 30%, 20%.

Maybe negative values could be the fixed-width columns ( e.g. -40 would mean this segment has a fixed width of 40 Fyne-pixely-units ). Kind of confusing and might be trying to tack too much on to this layout as I mentioned above.

andydotxyz:
We don't really go for fixed width so it is hard to know how to go ahead. But if the portion is "0" then the `MinSize` would be respected I would guess?

Jacalz:
That seems like a good plan. I'll try to implement all of what we have discussed here when I have some time over. Been quite down and out the last couple days (and still recovering as of now) with a flu.

andydotxyz:
> That seems like a good plan. I'll try to implement all of what we have discussed here when I have some time over. Been quite down and out the last couple days as I am currently recovering from a flu.

Oh no, rest up!

andydotxyz:
Just checking that the update is desired - it no longer is possible to do what is highlighted at the top:

> The example below contains four buttons that are set to take up 30%, 20%, 30% and 10% respectively (with remaining 10% empty at the end).

I think when working with sum it will no longer leave space. I'm assuming this is desired, but I wanted to be sure.

Jacalz:
Yeah it is to be coerced. I just haven't had time to look at this more than the drive by commit. It seemed like people preferred to have it do the sum instead of just predefined portions. That option also seems more robust (and I hope that it might be less prone to rounding errors as well) :)

andydotxyz (approved):
Cool, thanks

Jacalz:
Thanks for the approval but I intend to add tests before merging. I'm in too deep and can't break my New Year's resolution ;)

andydotxyz (approved):
Nice :)

PR #65, opened 2023-06-19
Add Adwaita theme

metal3d:
The color scheme is generated with:

`go generate ./theme`

It fetches the palette from Adwaita documentation page.


andydotxyz (changes requested):
Great stuff thanks.

However some of the colour mapping isn't right. The colour names (blue/green etc) are just used where user may desire a different primary colour (see fyne_settings). So you should map instead to the semantic names primary/success/warning/error, which is pretty much 1:1 match.

Or both, to provide consistency for both areas.

metal3d:
I need to add Green, Red, Brown... colors... That would be more consistent. Let me a few minutes to add them.

metal3d:
@andydotxyz I'm not sure if you've seen I made requested changes 😉

andydotxyz:
> @andydotxyz I'm not sure if you've seen I made requested changes 😉

Apologies I did miss this - been a busy week!

andydotxyz (approved):
This is great, thanks so much!

Issue #54, opened 2023-03-18
Mark selected date in calendar


Dmitriy147:
Is it possible to make selected day in calendar with another text color?

andydotxyz:
Not yet, this sounds like a good feature to add.

Dmitriy147:
Something like this..![Screenshot_20230319-114255.jpg](https://user-images.githubusercontent.com/78656465/226163611-3910cde2-0521-445e-810a-55bb4ca378b9.jpg)

Dmitriy147:
Done!

andydotxyz:
Are you sure this was done? I can't see the code to handle it

Dmitriy147:
I need some time to get acquainted with the git-fork and make changes..

Jacalz:
Let’s keep this open then until the code lands on a branch in this repository.

Dmitriy147:
Created, finaly, my variant of calendar...
https://github.com/Dmitriy147/fynecalendar

PR #66, opened 2023-06-22
Add different selection modes to Calendar (Single, Multi, Range)

PucklaJ:
This PR adds different modes of selection to the calendar.

+ CalendarSingle
This is the same as the current mode. Only one date can be selected at once.
+ CalendarMulti
Multiple random dates can be selected at the same time
+ CalendarRange
Select a beginning and end to select all dates in between those dates.

To achieve this change some compatibility breaking changes had to be made:

1. `SelectedDates` field has beend added to the calendar struct
Here all dates that the user will select are stored no matter what mode has been chosen. This can also be freely manipulated as it is an exported field. When calling `Refresh` the display is updated.
2. ~~`OnSelected` has been replaced by `OnChanged`~~ `OnChanged` has been added as a callback for the new modes
The `OnChanged` callback is now called everytime the user presses a button and the selected dates are changed. While `OnSelected` is called with the clicked date if the date is part of the selected dates.
3. Selected date buttons are highlighted using the HighImportance
Since there isn't another way to give buttons custom colors right now. I used the Importance system to signal the user which dates have been selected. And if a button is not selected it uses the previous LowImportance.

This also takes care of #54

### DIFF
--- a/cmd/calendar_demo/main.go
+++ b/cmd/calendar_demo/main.go
@@ -15,15 +15,17 @@ func main() {
 	a := app.New()
 	w := a.NewWindow("Calendar")
 
-	i := widget.NewLabel("Please Choose a Date")
+	i := widget.NewLabel("OnChanged")
 	i.Alignment = fyne.TextAlignCenter
 	l := widget.NewLabel("")
 	l.Alignment = fyne.TextAlignCenter
-	d := &date{instruction: i, dateChosen: l}
+	ll := widget.NewLabel("OnSelected: ")
+	d := &date{selectedDates: l, dateSelected: ll}
 
 	// Defines which date you would like the calendar to start
 	startingDate := time.Now()
-	calendar := xwidget.NewCalendar(startingDate, d.onChanged)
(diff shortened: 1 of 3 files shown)


andydotxyz (changes requested):
I would expect this could be made backward compatible by defaulting to single data selection. Then a field or method could change this after constructor returns?

PucklaJ:
I changed the constructor to default to single mode. It is more backwards compatible, but not completely since the callback requires a slice instead of a single variable.

andydotxyz:
Perhaps `OnSelected` is still valid, as is. And a new callback (OnChanged?) might provide the ability for the slice/range types of callback?

PucklaJ:
@andydotxyz I readded the OnSelected callback and changed the default constructor to the way it was before. For the Single mode it works the same way and for the Multi and Range mode I called the callback everytime a date is pressed and part of the selected dates.

andydotxyz:
I'm very sorry this PR got left for a while without re-review.
The `Calendar` has since been promoted to the `fyne` repo on develop branch - would you be interested in re-creating the work over there. We are working on some time input widgets that provide a single "input field" type approach which might make it easier to see how multi-select calendar fits in?

PR #67, opened 2023-06-25
Added date_input_widget

heejit:


Jacalz:
Can you please stick to one item per PR? I don’t we should land separate widgets and layouts within the same PR.

Is the HBox layout not kind of a duplicate of https://github.com/fyne-io/fyne-x/pull/64?

heejit:
I wanted to layout widget horizontally and can't find any other layout if it is same then reject it.

heejit:
I will stick to one item per PR for now on.

Jacalz:
In that case, can you please update the PR (or open a new one id you prefer) that only includes one of the features?

I do think that the layout in question indeed is mostly the same as the one linked above. I'll try it complete that PR some time in the future. The full list of layouts in Fyne can be found here: https://developer.fyne.io/explore/layouts

heejit:
okay let me just add the date input widget

heejit:
I have remove the layout code.

Jacalz:
Thanks. I'll try to review the code some day going forward

widget/widget_date_input.go
```diff
@@ -0,0 +1,195 @@
...
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/widget"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const ConstDefaultDateFormat = "02-Jan-2006"
+
+var DayPos = []int{0, 2}
```
andydotxyz, line 14: Should these fields be part of the public API?

widget/widget_date_input.go
```diff
@@ -0,0 +1,195 @@
...
+var MonthPos = []int{3, 6}
+var YearPos = []int{7, 11}
+
+type MyDateEntry struct {
```
andydotxyz, line 18: I think naming here could be improved

widget/widget_date_input.go
```diff
@@ -0,0 +1,195 @@
...
+func (e *MyDateEntry) FocusLost() {
+	e.parse_and_update_date()
+	e.Entry.FocusLost()
+}
+
+// Date string to time.Time conversion
+// we assume 1st part is always Day
+// input = 1 -> 1-CurMonth-CurYear
+// input = 1.5, 1/5, 1-5 -> 1-5-CurYear
+func (e *MyDateEntry) parse_and_update_date() {
```
andydotxyz, line 139: Try to match the Go style for naming - your other methods are formatted correctly. `parseAndUpdateData` would be more in-keeping

andydotxyz (changes requested):
I just had a quick glance over and a few things stood out.
Also can you add an item to the README with an image to show it in action please?

widget/widget_date_input.go
heejit: No

widget/widget_date_input.go
heejit: okay

widget/widget_date_input.go
heejit: okay

widget/widget_date_input.go
heejit: I will update naming as per go naming rule

andydotxyz:
I just came back to this (sorry for the delay) and notice that `MyDateEntry` is still a fairly strange name ;).
Also if you can rebase onto latest `main` the tests should all pass again.

heejit:
> I just came back to this (sorry for the delay) and notice that `MyDateEntry` is still a fairly strange name ;). Also if you can rebase onto latest `main` the tests should all pass again.

I have rewrite the whole widget this time using /richtext : https://github.com/heejit/fyne_codes/blob/main/widget_date_input.go

widget/widget_date_input.go
```diff
@@ -0,0 +1,208 @@
...
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+	"time"
+)
+
+var _ fyne.Focusable = (*JDateInputWidget)(nil)
+var _ fyne.Tappable = (*JDateInputWidget)(nil)
+
+// Custom Date Input Widget using RichText
+type JDateInputWidget struct {
```
Jacalz, line 16: Someone using this will get `widget.JDateInputWidget` which stutters. No need to say "widget" again. Also, what is the "J" for?

The README mentions `DateEntry` so why not call it that?

widget/widget_date_input.go
```diff
@@ -0,0 +1,208 @@
...
+	t.border.FillColor = theme.BackgroundColor()
+	t.border.StrokeColor = theme.InputBorderColor()
+	t.daySeg.Style.TextStyle.Bold = false
+	t.monthSeg.Style.TextStyle.Bold = false
+	t.yearSeg.Style.TextStyle.Bold = false
+	t.Refresh()
+}
+
+func (t *JDateInputWidget) TypedRune(k rune) {
+	// needed
```
Jacalz, line 104: Not needed? I know it is needed for the interface but the comment makes that unclear. If you don't need this method for anything other than satisfying the interface, you can just remove the comment

widget/widget_date_input.go
```diff
@@ -0,0 +1,208 @@
...
+		t.updateDisplay()
+	} else if e.Name == fyne.KeyPageUp {
+		t.addTime("d", 10)
+		t.updateDisplay()
+	}
+}
+
+// End: implement focusable interface
+
+// Start: implemnet Tappable
```
Jacalz, line 146: Should be "implement". This comment applies to more places

Jacalz (changes requested):
Thanks for the update. I just did a very quick code inspection on my phone and probably missed some things but I hope the review is helpful either way.

widget/widget_date_input.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// widget render
+func (t *JDateInputWidget) CreateRenderer() fyne.WidgetRenderer {
+	t.ExtendBaseWidget(t)
+	return widget.NewSimpleRenderer(t.gl)
+}
+
+// constructor
+func NewJDateInputWidget(app fyne.App, window fyne.Window) *JDateInputWidget {
```
Jacalz, line 181: All exported functions and methods should have a comment describing if them. See already existing code in this repo for examples.

Jacalz (changes requested):

widget/widget_date_input.go
heejit: I want to prefix my widget names with J no special meaning

Jacalz: We need our APIs to be consistent and be easy to understand. I would suggest `DateInput`.

widget/widget_date_input.go
Jacalz: The comments also need to follow the Go standard starting with the name of the function.

andydotxyz:
I just wanted to see the status of this and tried to resolve the conflict but as it was a PR from your master branch GitHub warned me not to, so I didn't - but it's a simple change.

I am sorry it took so long to get back to this - I hadn't noticed that there were commits since the last feedback.
However the naming is still not resolved as requested by Jacalz and there are references to your fork in the README.

Also it seems like you're looking to behave like an Entry but it isn't an Entry - so why not just extend Entry?
I have been working on an alternative picker based on the `Calendar` here and a simple Entry widget with an accessory for popping up the calendar, changes to the Entry theming made this fairly simple - not sure if this changes your thoughts?

Screenshot 2024-09-07 at 18 30 15

heejit:
I wanted dateinputwidget to used via keyboard(arrow keys) that is why designed like that and if you already working on it then there is no need of this again.

andydotxyz:
The keyboard input could be a great extension of what has landed in `fyne` - if you would consider basing such a feature on top of the new widget that would be amazing, thanks.
However closing this PR so we can focus over there now.

heejit:
okay

PR #69, opened 2023-07-03
Fix/hexlayout

andydotxyz:
Updates to work better with the theme padding and line fixes in the Fyne version recently required from fyne-x updates.

Now looks like this:

Screenshot 2023-07-03 at 13 00 34


andydotxyz:
No requests for change in 1 week. This is minor tweaks, so merging as per default.

Issue #39, opened 2022-06-09
Responsive breaks TextWrapWord of widget.Label


hennedo:
It seems the responsive layout breaks the TextWordWrap of widget.Label.
If I build the example [cmd](https://github.com/fyne-io/fyne-x/blob/master/cmd/responsive_layout/main.go) with go 1.18.1 on MacOS (ARM) it does not wrap the text if the window gets smaller. It only wraps the text where there is an \n in the code.

metal3d:
Right, and not only this.

Buttons are not clickable.

I'm checking what's wrong.

metal3d:
I think I found the problem - sorry for the delay. Waiting for review and tests in #70

metal3d:
Fixed in #70 - you can test and close this issue if possible.

Thanks a lot for your bug report !

PR #70, opened 2023-07-05
Fixes renderer creation on ResponsiveLayout

metal3d:
Issue #39 - calling the inside object `CreateRenderer` breaks several native behaviors (text wrapping, clickable, and many others).

Removing this call and using NewSimpleRenderer on the reponsiveWidget fixes the problems.

I don't have precise explanation.


andydotxyz (approved):

andydotxyz:
Yes this makes sense - you are accidentally assigning a renderer to two different objects I think... 
anyhow glad it is sorted, this looks better.

PR #71, opened 2023-07-08
Add HBoxCusomPadding and VBoxCustomPadding layouts

dweymouth:
This PR adds two new layouts, HBoxCustomPadding and VBoxCustomPadding. The code is adapted from fyne's BoxLayout with the change that a custom amount of padding, `ExtraPad`, can be added to the default theme padding.

### DIFF
--- a/layout/hboxcustompadding.go
+++ b/layout/hboxcustompadding.go
@@ -1,0 +1,94 @@
+package layout
+
+import (
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/layout"
+	"fyne.io/fyne/v2/theme"
+)
+
+var _ fyne.Layout = (*HBoxCustomPadding)(nil)
+
+// HBoxCustomPadding is a layout based on Fyne's HBox layout that
+// allows for the amount of padding between items to be customized.
+// The padding between items is theme.Padding() + ExtraPad (which may be negative).
+// If DisableThemePad is set, the padding is just ExtraPad. This is
+// useful for placing items that directly touch each other (with ExtraPad=0).
+type HBoxCustomPadding struct {
+	DisableThemePad bool
+	ExtraPad        float32
+}
+
+func (*HBoxCustomPadding) isSpacer(obj fyne.CanvasObject) bool {
+	if !obj.Visible() {
+		return false
+	}
+	if spacer, ok := obj.(layout.SpacerObject); ok {
+		return spacer.ExpandHorizontal()
+	}
+
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


andydotxyz:
Why would you need `DisableThemePad` plus `ExtraPad`? Why not just `Padding`?
If developers wanted "default padding" they would use `VBox` or `HBox`

dweymouth:
> Why would you need `DisableThemePad` plus `ExtraPad`? Why not just `Padding`? If developers wanted "default padding" they would use `VBox` or `HBox`

My use case is mostly placing things a bit closer together than they otherwise would be, with a negative `ExtraPad`. I guess you're right that just having `Padding` is a simpler API and in that case I could just initialize it with `Padding: theme.Padding() - x`. Only downside I guess is that if the theme padding updates you need to know about it and update all the *BoxCustomPadding layouts yourself. (But I don't expose a way to change theme padding in my app)

dweymouth:
> > Only downside I guess is that if the theme padding updates you need to know about it and update all the *BoxCustomPadding layouts yourself. (But I don't expose a way to change theme padding in my app)
> 
> Isn't this equally true for both designs?

As-is, it calls `theme.Padding()` every layout, so if the value changes the layout will automatically update

andydotxyz:
Ah I understand - I thought you had meant the standard padding changing between releases.

I guess I'd think with a name like "custom padding" it would not be derived from the standard one anyway, but it's not my use-case so feel free to correct me.

dweymouth:
I originally added it without the DisableThemePad at all, and my use case was just to tweak the spacing a bit in certain cases, so I wanted to work with the theme padding value and do things like "move these VBoxed labels 5 units closer together". I added DisableThemePad later when I wanted to use it to pack buttons together touching each other to use as a segmented control.

My use case is fully possible with the API of `Padding` only instead of `ExtraPad` + `DisableThemePad` (since I don't worry about theme.Padding() changing during the lifetime of the app running)- and indeed it is a simpler API. I wonder what @Jacalz or others think

Jacalz:
I think the PR seems sensible the way it is set up at the moment. Just having a `Padding` field seems like it would become problematic as soon as your load a custom theme.

andydotxyz:
> I think the PR seems sensible the way it is set up at the moment. Just having a `Padding` field seems like it would become problematic as soon as your load a custom theme.

That is true - but the inclusion of `DisableThemePad` is the same as returning `myPadAmount - theme.Padding()` so I'm not sure it really warrants another API. Alternatively this PR could be always disabling the default and you could add as `myPadAmount + theme.Padding()`. With this bool it feels like it is both and cannot decide - is that really needed?

PR #79, opened 2023-10-30
Fix responsive

metal3d:
- The padding were not respected between element in a line and stepping
  to the next line
- We now prefer move the first element in a row to the X padding, to
  avoid computation - the responsive behavior is to center elements in a
  row, so it must be OK
- We accept now that a non-responsive element is added to the container,
  it will be resized to the default 100% of the container width
- We previously use the window size to compute the reponsive element
  sizes. That's an error... The reponsivity should be set to the
  container size. That fixed now.

### DIFF
--- a/layout/responsive.go
+++ b/layout/responsive.go
@@ -28,7 +28,17 @@ import (
 // responsiveBreakpoint is a integer representing a breakpoint size as defined in Bootstrap.
 //
 // See: https://getbootstrap.com/docs/4.0/layout/overview/#responsive-breakpoints
-type responsiveBreakpoint uint16
+type responsiveBreakpoint = float32
+
+const (
+	Full     float32 = 1.0
+	Half     float32 = 0.5
+	OneThird float32 = 1.0 / 3.0
+	TwoThird float32 = 2.0 / 3.0
+	Quarter  float32 = 0.25
+	Fifth    float32 = 0.2
+	Sixth    float32 = 1.0 / 6.0
+)
 
 const (
 	// SMALL is the smallest breakpoint (mobile vertical).
@@ -110,24 +120,15 @@ func (resp *ResponsiveLayout) Layout(objects []fyne.CanvasObject, containerSize 
 		return
 	}
 
-	// Responsive is based on the window size, so we need to get it
-	window := fyne.CurrentApp().Driver().CanvasForObject(objects[0])
-	if window == nil {
-		return
-	}
-
 	// this will be updatad for each element to know where to place
 	// the next object.
-	pos := fyne.NewPos(0, 0)
+	pos := fyne.NewPos(theme.Padding(), 0)
 
 	// to calculate the next pos.Y when a new line is needed
-	maxHeight := float32(0)
+	var maxHeight float32
 
 	// objects in a line
 	line := []fyne.CanvasObject{}
-
-	// cast windowSize.Width to responsiveBreakpoint (uint16)
-	ww := responsiveBreakpoint(window.Size().Width)
 
 	// For each object, place it at the right position (pos) and resize it.
 	for _, o := range objects {
@@ -138,7 +139,9 @@ func (resp *ResponsiveLayout) Layout(objects []fyne.CanvasObject, containerSize 
 		// get tht configuration
 		ro, ok := o.(*responsiveWidget)
 		if !ok {
-			log.Fatal("A non responsive object has been packed inside a ResponsibleLayout. This is impossible.")
+			// We now allow non responsive objects to be packed inside a ResponsiveLayout. The
+			// size of the object will be 100% of the container.
+			ro = Responsive(o).(*responsiveWidget)
 		}
 		conf := ro.responsiveConfig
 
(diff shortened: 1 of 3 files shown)


Jacalz:
You can delete the old one entirely if you want to. We have no stability guarantees in this repo :)

metal3d:
> You can delete the old one entirely if you want to. We have no stability guarantees in this repo :)

This is the case 🙂

I use the layout as I should and I added the container object.

metal3d:
Oh I forget to change the layout.Responsive inside the demo.
I'll do it in a few minutes.

container/responsive_ratios.go
```diff
@@ -0,0 +1,27 @@
+package container
+
+// Ratio is a helper type to define a fraction of a container. It's a float32 alias.
+type Ratio = float32
+
+// Some helpers to define common ratios.
+
+// FullRatio is the full size of the container.
+func FullRatio() Ratio { return 1.0 }
```
aodhan-domhnaill, line 9: Why not define constants?

container/responsive.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// Example:
+//
+//	NewResponsive(
+//		widget.NewLabel("Hello World"), // 100% by default
+//		Responsive(widget.NewLabel("Hello World"), 1, .5), // 100% for small, 50% for others
+//		Responsive(widget.NewLabel("Hello World"), 1, .5), // 100% for small, 50% for others
+//	)
+func NewResponsive(objects ...fyne.CanvasObject) *fyne.Container {
+	container := container.New(layout.NewResponsiveLayout())
```
andydotxyz, line 22: You can pass the objects in the constructor

container/responsive.go
```diff
@@ -0,0 +1,59 @@
...
+//		Responsive(widget.NewLabel("Hello World"), 1, .5),
+//		Responsive(widget.NewLabel("Hello World"), 1, .5, .25),
+//	)
+//
+// Or with the Add() method:
+//
+//	ctn := NewResponsive()
+//	ctn.Add(Responsive(widget.NewLabel("Hello World"), 1, .5))
+//	ctn.Add(Responsive(widget.NewLabel("Hello World"), 1, .5))
+func Responsive(object fyne.CanvasObject, ratios ...Ratio) fyne.CanvasObject {
```
andydotxyz, line 57: Hmm, is adding more static functions desirable to handle what is metadata for a single type of container?
It may be a wider design question but if the info cannot be passed in constructor is there a simpler way to handle it that keeps the API entry points to a lower number?

My concern is that "Responsive" could be assumed to work in other places. Like if we accidentally added a "StayLeft" function to describe the position of an item in a border container...

layout/responsive.go
```diff
@@ -239,30 +164,37 @@ func (resp *ResponsiveLayout) maxFloat32(a, b float32) float32 {
...
 	r := &ResponsiveLayout{}
 
-	objects := []fyne.CanvasObject{}
-	for _, unknowObject := range o {
-		if _, ok := unknowObject.(*responsiveWidget); !ok {
-			unknowObject = Responsive(unknowObject)
-		}
-		objects = append(objects, unknowObject)
-	}
+	//objects := []fyne.CanvasObject{}
```
andydotxyz, line 172: Probably should remove the dead code

andydotxyz (commented):
Lots of nice improvements in there, thanks for submitting this. However I'm not sure that adding more "Responsive" entry points is desirable, is there a way to have fewer package global API with this change? Like if the container had Ali for managing ratios instead of passing in wrapped types?

container/responsive.go
metal3d: It's a question I've been asking myself for some time.

`NewReponsive` returns a `*fyne.Container` where the `Add()` method only accepts `fyne.CanvasObject`, which is perfectly fine.

As a result, I have two possible choices:

- either I allow metadata to be added, as I do with `Reponsive()`, but this effectively transforms the objects into "something else with metadata". But the advantage is that I can always add them to another container without affecting the way it works (for example, to test, I change my container to Grid, which doesn't cause any problems).
- or I overload `fyne.Container` to return a `ResponsiveContainer` that have a `Add()` method that accepts responsive configuration. But then I'll have two problems:
    - if I change my container to test another one, the "Add()" method will no longer be respected
    - functions that accept `fyne.Container` will refuse my object

I was thinking of allowing this with `NewResponsive()` by returning a "config" function signed like this: `func(widget, ratios... float32)`.

So to do it this way:

```go
ctn, configFunc := NewResponsive(label, entry)

configFunc(label, 0.5, 1)
configFunc(entry, 1)

// But how do I add a widget now?
ctn.Add(label2) // This may refresh the view... 
configFunc(label2, 0.5) // and here too
```
It doesn't look very attractive to me.

container/responsive_ratios.go
metal3d: To not store the value. Functions are compiled only if they are used. It's, it seems, a good practice. Maybe I'm wrong.

aodhan-domhnaill: Constants are the same https://stackoverflow.com/a/38982889 Most Go I see uses constants

metal3d: Yes.

But for example, Fyne returns colors from function calls, like the Padding or like some several values that are "constant". I propose functions to ensure that we can adapt the ratios from possible futur configuration. For example, following themes.

If it's really a problem, I can of course change the code and set it as constants.

container/responsive.go
metal3d: OK, I think we can adapt the `NewResponsive` function to make this:

```go

label := widget.NewLabel...
entry := widget.NewEntry...
other := widget.NewEntry...

config := []Ratios{
  {Object: label, Ratios: []float32{1, 0.5}}, // only the label
}

// use the config (or nil)
ctn := NewReponsive(&config, label, entry) // note that entry is not configured

// adding anthor widget
config.SetRatio(other,0.5, 0.5, 1) // if needed...
ctn.Add(other)
```

That means that I need to keep the `Ratios` configuration inside the layout, and to remove the `Responsive()` function + responsive widget.

andydotxyz: This seems cleaner

container/responsive_ratios.go
andydotxyz: The reason behind theme returning values from a function is that "primary" colour could be respresented by any colour - changed at runtime. However isn't the concept of "TwoThird" always going to be 2/3? If the names were somehow semantic the discussion would track but their names do indicate staticness...

andydotxyz (commented):
I am so sorry this got dropped for such a long time. Are you still happy to refine this to land? I think the discussion about constant named functions returning constant values is an interesting discussion to just complete and see if anything needs to change.

metal3d:
I'm of course always open to make modifications.
Even if I'm a bit busy for months, I still follow what happens on Fyne and FyneX 🙂

I will take some times to refine my code and my observation.

PR #87, opened 2024-03-25
Add about dialog and window for helping apps make a pretty about screen

andydotxyz:
Matching the parallax look in develop version of fyne_demo.
Looks sweet :)

about

also added a new general “fyne-x” demo command for when we don’t really need a full app to show off a new component.


Jacalz:
Nice stuff. Kind of related to https://github.com/Jacalz/rymdport/issues/143 which I opened recently.

Something about having the version above seems strange to me. Maybe we can consider something like the following screenshot (smaller text below the icon)?

dialog/about.go
```diff
@@ -0,0 +1,136 @@
...
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/dialog"
+	"fyne.io/fyne/v2/layout"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+// ShowAbout opens a parallax about dialog using the app metadata along with the
+// markdown content and links passed into this method.
+func ShowAbout(content string, links []*widget.Hyperlink, a fyne.App, w fyne.Window) {
```
Jacalz, line 17: Can we get a `NewAbout()` so it is possible to not have to recreate it each time?

dialog/about.go
```diff
@@ -0,0 +1,136 @@
...
+// ShowAboutWindow opens a parallax about window using the app metadata along with the
+// markdown content and links passed into this method.
+func ShowAboutWindow(content string, links []*widget.Hyperlink, a fyne.App) {
+	w := a.NewWindow("About")
+	w.SetContent(aboutContent(content, links, a))
+	w.Resize(fyne.NewSize(360, 280))
+	w.Show()
+}
+
+func aboutContent(content string, links []*widget.Hyperlink, a fyne.App) fyne.CanvasObject {
```
Jacalz, line 32: This function is huge. Can you make it a bit smaller and make readable?

dialog/about.go
```diff
@@ -0,0 +1,136 @@
...
+
+type unpad struct {
+	top, bottom bool
+}
+
+func (u unpad) Layout(objs []fyne.CanvasObject, s fyne.Size) {
+	pad := theme.Padding()
+	var pos fyne.Position
+	if u.top {
+		pos = fyne.NewPos(0, -pad)
```
Jacalz, line 119:
```suggestion
		pos.Y = -pad
```

dialog/about.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	size := s
+	if u.top {
+		size = size.AddWidthHeight(0, pad)
+	}
+	if u.bottom {
+		size = size.AddWidthHeight(0, pad)
+	}
```
Jacalz, line 127:
```suggestion
	if u.top {
		size.Y += pad
	}
	if u.bottom {
		size.Y += pad
	}
```

Jacalz (changes requested):

andydotxyz:
That's all addressed now (and new images added to show)

Jacalz (approved):
Thanks. This is looking perfect now. I haven't been able to test it locally (due to being away from my laptop) but the code looks great.

PR #95, opened 2024-12-05
fix(completionentry): calculate max size and resize correctly 

geodimm:
The new max size calculation ensures the popup position is always below
the entry widget.

The addition of the Resize method ensures the popup is updated with the
rest of the widget upon resize.

Since now the popup doesn't go above the entry widget not all options
are rendered and the test had to be updated to select the first item as
the other ones are not visible

Before:
https://github.com/user-attachments/assets/baff9482-3543-4e8a-8e52-cdad1ef79dcc

After:
https://github.com/user-attachments/assets/981f6a41-7705-42d0-a075-555c65fc28df

### DIFF
--- a/widget/completionentry.go
+++ b/widget/completionentry.go
@@ -146,8 +146,7 @@ type navigableList struct {
 }
 
 func newNavigableList(items []string, entry *widget.Entry, setTextFromMenu func(string), hide func(),
-	create func() fyne.CanvasObject, update func(id widget.ListItemID, object fyne.CanvasObject),
-) *navigableList {
+	create func() fyne.CanvasObject, update func(id widget.ListItemID, object fyne.CanvasObject)) *navigableList {
 	n := &navigableList{
 		entry:           entry,
 		selected:        -1,
--- a/widget/completionentry.go
+++ b/widget/completionentry.go
@@ -52,6 +52,15 @@ func (c *CompletionEntry) Refresh() {
 	}
 }
 
+// Resize sets a new size for a widget.
+// Note this should not be used if the widget is being managed by a Layout within a Container.
+func (c *CompletionEntry) Resize(size fyne.Size) {
+	c.Entry.Resize(size)
+	if c.popupMenu != nil {
+		c.popupMenu.Resize(c.maxSize())
+	}
+}
+
 // SetOptions set the completion list with itemList and update the view.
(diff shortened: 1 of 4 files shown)


widget/completionentry.go
```diff
@@ -136,7 +146,8 @@ type navigableList struct {
 }
 
 func newNavigableList(items []string, entry *widget.Entry, setTextFromMenu func(string), hide func(),
-	create func() fyne.CanvasObject, update func(id widget.ListItemID, object fyne.CanvasObject)) *navigableList {
+	create func() fyne.CanvasObject, update func(id widget.ListItemID, object fyne.CanvasObject),
```
andydotxyz, line 149: I don't understand this change - it looks like you have added another item to the parameters to put it on a new line?

andydotxyz (commented):
Thanks for looking at this, just asking something that looks strange in the style

widget/completionentry.go
geodimm: It's just my formatter automatically formatting the source code, it's a non-functional change :)

geodimm: @andydotxyz Would you prefer to undo the white space change or you're happy to merge the PR as is?

andydotxyz: Yes, please do put it back to not having an empty parameter which is just newline.
I will get our CI updated so that the linter runs correctly for our code standards.

geodimm: Done :)

geodimm: @andydotxyz would it be possible to get this merged soon? 🙏

andydotxyz (approved):
Perfect thanks so much for this, and congratulations on your first fyne-x PR landing :)

andydotxyz:
I wonder if there maybe should be a follow-up for when the select is at the bottom of a screen and should pop above instead of below? Perhaps there should be a minimum size check and if the "down" is too short use "up" instead?

geodimm:
Yes, I was also thinking about making the size configurable, e.g. for the dropdown to take 50% or 75% of the screen size as opposed to filling it. Both would be good improvements

andydotxyz:
A word of warning - don't make widget size relative to screen/window size - it leads to bad outcomes. If you want to configure how much content should be displayed then it should be relative to some content reasoning not window space.

PR #97, opened 2024-12-29
BottomUpVBoxLayout for mobile devices

BieHDC:
I made and use this for one mobile app at this point that is being used with one hand only.
In Fyne everything aligns by top-left by default and this makes it bottom-left.
Screenshots show what i mean.


andydotxyz:
You could get the same effect with a standard VBox by putting a Layout.Spacer at the first item. I don't think we need a custom layout for this purpose.

BieHDC:
> You could get the same effect with a standard VBox by putting a Layout.Spacer at the first item. 

I didnt know that was a thing, however [that explanation](https://docs.fyne.io/api/v2.5/layout/spacer.html) doesnt really give away what it does for me: `Spacer is any simple object that can be used in a box layout to space out child objects`.

Closing as solved.

PR #98, opened 2024-12-31
DynamicGridLayout - Initial Draft

BieHDC:
This is a revised version of the dynamic grid i used in [fic](https://github.com/BieHDC/fic) to display a preview of a folders's images. It takes the available space and tries to fit as many items as possible and if it can fit them all it tries to stretch them out over the available space. So if a folder has only few images, they are as large as possible, and if it has a lot of images it will squeeze as much as possible by checking each MinSize and scaling everything by the largest encountered size.

There are various fixme-s in the code that need addressing, however a general discussion beforehand on how this layout should work takes priority. In the Video you see one of those bugs which in real world use you likely wont encounter, because you would fill it with pretty much same sized widgets, but it would be a hard blocker for me to release.

And right now the code is in main because of heavy development. If this gets accepted and finalised, it goes into the layout folder of course.

https://github.com/user-attachments/assets/42c22f63-404a-4d74-a1d4-0680b179cf0a

### DIFF
--- a/cmd/dynamicgrid_demo/main.go
+++ b/cmd/dynamicgrid_demo/main.go
@@ -1,0 +1,268 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"math"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/app"
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/theme"
+	"fyne.io/fyne/v2/widget"
+)
+
+func main() {
+	a := app.New()
+	w := a.NewWindow("Dynamic Grid")
+
+	const numTestItems = 128
+	names := make([]string, 0, numTestItems)
+	for i := 0; i < numTestItems; i++ {
+		name := fmt.Sprintf("Button %d", i)
+		names = append(names, name)
+	}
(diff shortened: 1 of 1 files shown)


Jacalz:
How is this different from `widget.GridWrap` upstream in Fyne?

BieHDC:
> How is this different from `widget.GridWrap` upstream in Fyne?

It does not make a scroll container, it only uses the available space.
When there is more space than elements, it increases the sizes of the elements to fill as much space as possible.
It puts the user directly into control on which widgets are being up into it instead of the length, create and updateitem strategy.
And that also allows for a easier method of selecting which widgets are being embedded.
There is no and there is not meant to be an inner item select handling. The items handle themselves.

Those are the main points.

Jacalz (changes requested):
Thanks for the clarification. That sounds like a different container indeed. Would you mind creating a container package here and adding it as an importable type? Right now it is just a runnable demo and no container that can be used by other people.

Jacalz:
Ah, sorry. I read now that you wanted a discussion about it. Let's have that before. Disregard my previous comment.

andydotxyz:
> Would you mind creating a container package here and adding it as an importable type?

It could also just be a layout as @BieHDC implied above.

BieHDC:
Once the `fixme`s in the code have been answered i will turn it into a proper package.

PR #100, opened 2025-01-06
Modernize and use Go 1.19 as base

Jacalz:
Updates various dependencies and workflows. Now using Go 1.19 as base just like upstream Fyne.


andydotxyz (approved):

Jacalz:
The required checks need to be updated for this to be mergeable. The good news is that the highest version of Go used now names the workflows as "stable" so they will never need to be changed again.

andydotxyz:
All sorted :)

Jacalz:
Great :)

PR #101, opened 2025-01-21
Spinner Widget

jimorc:
Spinner widget displays a integer and  buttons for incrementing and decrementing the integer value.
The spinner has a minimum and maximum value, as well as a step value (the amount to increment or decrement by.
The initial value is set either to the minimum value, or to that of a bound integer.

It is possible to set the minimum, maximum, and step values either when the widget is created, or later via a call to `Spinner.SetMinMaxStep`. In either case, the values are validated as follows:
* max > min
* step > 0
* step <= max - min
The increment and decrement buttons are enabled/disabled based on the spinner value. For example:
* if the value = min, then the decrement button is disabled.
* if the value = max, then the increment button is disabled.
The spinner value may be set by:
* Clicking the increment/decrement buttons.
* Changing the bound integer value.
* Calling `Spinner.SetValue`.
* Via the keyboard when the widget has focus:
    * KeyUp or '+' to increment the value.
    * KeyDown or '-' to decrement the value.
* Via the mouse scroller or via touchpad when the widget has focus.

If either the bound value or the value passed to SetValue is outside the range of min to max, the value is reset to be within the range. For example, if value < min, then the value is set to min; if value > max, then the value is set to max.

All the standard widget functionality is provided. For example:
* Enable/Disable
* FocusGained/FocusLost
* Bind/Unbind an integer value

The Spinner widget is tested and supported on the desktop only; functionality has not been provided for Android or iOS.
The code has been tested on Ubuntu 24.04, Windows 11, and MacOS 15.1


andydotxyz:
Could you include a screenshot of the new widget please?

andydotxyz:
> The Spinner widget is tested and supported on the desktop only; functionality has not been provided for Android or iOS.

Why do you think this won't work on mobile - it should be the same for all?

jimorc:
> > The Spinner widget is tested and supported on the desktop only; functionality has not been provided for Android or iOS.
> 
> Why do you think this won't work on mobile - it should be the same for all?

I would have to display a mobile keyboard. How and when should that be done: When the widget gains focus? Is there some other way to indicate that the keyboard should be displayed? I am not familiar with mobile programming.

jimorc:
This PR only implements an integer spinner. A float spinner could be useful as well. 
The simplest way to implement that is to use generics, but generics require go version 1.21 and fyne-x is set to use go version 1.19. What are the reasons for limiting fyne-x to go version 1.19?
Alternatively, it would be possible to have an IntegerSpinner, a Float32Spinner, and a Float64Spinner, but there would be a lot of almost duplicated code. When fyne-x is updated to version 1.21, then these could be deprecated in favor of a generic Spinner. Do we really want to do that?

jimorc:
This PR only implements an integer spinner. It would be possible to implement a float32 spinner, a float64 spinner, and other integer types as well by using generics. Should I do that?
I won't proceed with this until other aspects of this PR are deemed satisfactory. This PR should therefore be considered as a draft.

andydotxyz:
> I would have to display a mobile keyboard. How and when should that be done: When the widget gains focus? Is there some other way to indicate that the keyboard should be displayed? I am not familiar with mobile programming.

If the item is focusable then the keyboard will display.

You can also control the keyboard visibility using APIs of the mobile driver extension.

andydotxyz:
> This PR only implements an integer spinner. It would be possible to implement a float32 spinner, a float64 spinner, and other integer types as well by using generics. Should I do that?

I would think that a float64 would be good too. I would not bother with all the variations.

That would align with the data binding package where we have Int and Float types.

Widgets that have generic parameters in their constructors feel clunky to me and impede peoples learning due to the added complexity. Just the two types could of course wrap an inner generic version simplifying extension if you wanted later?

jimorc:
> I would think that a float64 would be good too. I would not bother with all the variations.
> 
> That would align with the data binding package where we have Int and Float types.
> 

I eventually came to the same conclusion.

> Widgets that have generic parameters in their constructors feel clunky to me and impede peoples learning due to the added complexity. Just the two types could of course wrap an inner generic version simplifying extension if you wanted later?

I tried creating a widget with generic parameters yesterday before you sent this message and eventually came to the same conclusion: very ugly and not very intuitive.

I attempted to create a generic parameter for binding.Int and binding.Float. I was unable to do so at the time, but with a bit more thinking, I might be able to do it with a change to the binding API. So, not worth it.

I will modify the PR to contain an IntSpinner and a Float64Spinner.

jimorc:
This PR is now ready for review and possible merge. I have taken it as far as I can. There is now an IntSpinner widget and a Float64Spinner widget rather than a Spinner widget.

jimorc:
NOW this is ready as far as I can tell. There is a single Spinner that contains float64 values and a print (display) format. If the format contains "%d" or "%+d", the value is converted to an integer for display; otherwise, the value remains as a floating point value and is displayed as specified by the format. Invalid formats result in an error display.

jimorc:
I am working on a completely new Spinner widget based on NumericalEntry. This PR will be closed and a new PR submitted when that work is done.

Issue #103, opened 2025-04-10
BUG: NumericalEntry does not allow the entry of negative numbers.


jimorc:
It is possible to set the initial value of a NumericalEntry widget to a negative value, but not to input a negative value. The NumericalEntry widget currently ignores the '-' character.

Modify NumericalEntry to accept the input of negative numbers.

jimorc:
Issue #104 has been merged. This fixes issue #103.

PR #104, opened 2025-04-10
Allow negative numbers to be input to NumericalEntry widget.

jimorc:
This implements fix for issue #103: NumericalEntry widget does not accept negative values (i.e. throws '-' character away).

### DIFF
--- a/widget/numerical_entry.go
+++ b/widget/numerical_entry.go
@@ -13,6 +13,8 @@ import (
 type NumericalEntry struct {
 	widget.Entry
 	AllowFloat bool
+	// AllowNegative determines if negative numbers can be entered.
+	AllowNegative bool
 }
 
 // NewNumericalEntry returns an extended entry that only allows numerical input.
@@ -36,7 +38,7 @@ func (e *NumericalEntry) TypedRune(r rune) {
 		return
 	}
 
-	if r == '-' && e.Entry.CursorRow == 0 && e.Entry.CursorColumn == 0 {
+	if r == '-' && e.AllowNegative && e.Entry.CursorRow == 0 && e.Entry.CursorColumn == 0 {
 		e.Entry.TypedRune(r)
 	}
 }
--- a/widget/numerical_entry.go
+++ b/widget/numerical_entry.go
@@ -28,9 +28,15 @@ func NewNumericalEntry() *NumericalEntry {
 //
 // Implements: fyne.Focusable
 func (e *NumericalEntry) TypedRune(r rune) {
-	if e.Entry.CursorColumn == 0 && e.Entry.CursorRow == 0 &&
-		len(e.Entry.Text) > 0 && e.Entry.Text[0] == '-' {
-		return
+	if e.Entry.CursorColumn == 0 && e.Entry.CursorRow == 0 {
+		if e.AllowNegative {
+			if len(e.Text) > 0 && e.Text[0] == '-' {
+				return
+			} else if r == '-' {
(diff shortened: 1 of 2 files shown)


jimorc:
@Jacalz I have made the requested change.

Since you are the original developer of NumericalEntry, what was the reasoning behind positive numbers only?

Jacalz:
Thanks. I don't think there was any specific reasoning. My use case required positive numbers and I just didn't consider someone wanting to use negative numbers :man_shrugging:

widget/numerical_entry.go
```diff
@@ -33,6 +35,11 @@ func (e *NumericalEntry) TypedRune(r rune) {
 
 	if e.AllowFloat && (r == '.' || r == ',') {
 		e.Entry.TypedRune(r)
+		return
+	}
+
+	if r == '-' && e.AllowNegative && e.Entry.CursorRow == 0 && e.Entry.CursorColumn == 0 {
```
Jacalz, line 41: Thinking about this, I wonder if we need to be smarter about how we allow writing negative signs. Right now you will be able to write multiple negative signs as long as you are at the first position. Is that something we should handle?

jimorc: I have updated the code to not allow input in front of '-' character.

jimorc: By the way, multiple '.' and ',' are allowed. I have a fix for that as well. Will do a PR after this one is accepted.

Jacalz: > By the way, multiple '.' and ',' are allowed. I have a fix for that as well. Will do a PR after this one is accepted.

Cool. You can put dots or commas to make large numbers more readable. I think the solution for that is a bit complicated.

widget/numerical_entry.go
```diff
@@ -26,13 +28,23 @@ func NewNumericalEntry() *NumericalEntry {
...
 	}
 
 	if e.AllowFloat && (r == '.' || r == ',') {
 		e.Entry.TypedRune(r)
+		return
+	}
+
+	if r == '-' && e.AllowNegative && e.Entry.CursorRow == 0 && e.Entry.CursorColumn == 0 {
+		e.Entry.TypedRune(r)
 	}
```
Jacalz, line 48: You can group the two new cases together to avoid some code duplication. They are more or less handling the same thing now. Something like this would be a bit cleaner I think (note that it is untested):

```go
if e.AllowNegative && e.Entry.CursorColumn == 0 && e.Entry.CursorRow == 0 {
  if len(e.Entry.Text) == 0 || len(e.Entry.Text) > 0 && e.Entry.Text[0] != '-' {
    e.Entry.OnTyped()
}
```

Jacalz (changes requested):
Thanks. Just one idea to clean up some code duplication

widget/numerical_entry.go
jimorc: Actually not very complicated. But should multiple dot/commas be allowed? If yes, then I will throw my "fix" away.

widget/numerical_entry.go
jimorc: I ended up needing three levels of ifs, but my changes do pass all tests.

Jacalz (approved):

Jacalz:
Thanks :)

widget/numerical_entry.go
Jacalz: I suppose the answer depends on what the strconv.ParseFloat function allows. We should likely match that

PR #105, opened 2025-04-10
Upgrade CI to maintain itself thus unbreak CI

Jacalz:
This adds in the changes that we have in most other repos as well. Things should update themselves, builds are cached for faster speed and Staticcheck supports Go 1.24 cleanly. Don't run benchmarks as part of test either.


andydotxyz (approved):
Thanks

Issue #106, opened 2025-04-12
Radix Separators, Thousands Separators, Parsing, and Validation in NumericalEntryommendation:__


jimorc:
### Problems

There are a number of problems with the current implementation of the NumericalEntry widget. They are numbered below (inside parentheses):

Currently, the NumericalEntry widget accepts the following:
- (minus) as the first character only.
- 0-9 to represent digits. (1)
- The full-stop (period or dot) and comma characters  as the possible radix (decimal) separator and the other as possible thousands separators. (2) Any number of these characters in any position  other than in front of the minus sign may be added to the string. (3)

Parsing of the value is left to the user (the programmer) of the widget. The main problem here is that strconv.ParseFloat appears to only support the full-stop character as the radix separator, and the text cannot contain thousands separators. (4)

None of this is documented anywhere. (5)

I propose that we deal with each of these problems as follows:

### Problem 1: Digits

Not all languages use the Arabic numerals (0 through 9). Support for non-Arabic numerals would be difficult because we would have to ensure that there is not a mixture of numerals (Arabic, non-Arabic for each language or locale). Note that various International standards only specify Arabic numerals.

__Recommendation:__ Only accept Arabic numerals, so no change. If non-Arabic number input is required, this would probably best be handled by creating a separate widget.

### Problem 2: Radix and Thousands Separators

The radix separator is the character used to separate the integer part of a number from its fractional part.
Various languages use either the full stop, interpunct (middle dot), or comma as the radix separator. Other languages, mostly those that accept digits other than Arabic numerals, also accept other language-specific characters as the radix separator.

__Recommendation:__ Accept the full stop, interpunct, and comma as the radix character.

The thousands separator is used to make the reading of large numbers easier. The separator character is typically positioned so as to separate each set of three digits, but this is not universal. In some countries, the thousands separator is used only in the integer portion of the number, while in others, it may also be used in the fractional part of the number. Each language, or even locale, may specify a different thousands separator. The thousands separator is usually one of the following:
- a full stop if the radix separator is a comma.
- a comma if the radix separator is a full stop or middle dot.
- a space, half-space, or thin-space character.
- an apostrophe.
- an underscore.
- a language- or locale-specific character.

[...]

PR #107, opened 2025-04-24
Localized NumericalEntry widget

jimorc:
This widget is an extension of the fyne Entry widget containing the functionality of the fyne-x NumericalEntry widget, but with the following enhancements:
1. Parses the system locale's minus sign, radix separator, and thousands separator runes, then accepts digits and these runes with their alternative runes. See the README file for more information.
2. It validates all input (typed key, `SetText`, `Append`, and clipboard paste to accept only allowed characters.
3. Allows binding to a float64.
4. Provides `SetValue` method to allow setting of the entry's value to a float64 value. Using this method, the user code does not have to generate the text equivalent to set the value in the entry
5. Provides `ParseFloat` method to retrieve a float64 value from the entry. Using this method, the user code does not have to worry about converting the entry's text to text that can be parsed by `strconv.ParseFloat`.
6. Provides a `ValidateText` validator that is tied to the entry's `Validator` field to validate the text in the entry. If the user requires more validation, for example when including this widget in a spinner widget, then the user can extend this entry widget and add a validator function or  method that call `ValidateText` before performing additional validation.

This widget is based on Entry rather than NumericalEntry because little of the NumericalEntry code is reusable in this new widget.

This issue completes issue #106, with changes to how it is addressed based on things I learned while implementing the request.

**Note to reviewer:**

I would suggest reviewing the README.md file before reviewing the code so that you get an idea of how a system's locale affects both the characters that are accepted by the widget and the validation.

### DIFF
--- a/widget/numerical_entry.go
+++ b/widget/numerical_entry.go
@@ -76,7 +76,7 @@ func (e *NumericalEntry) FocusLost() {
 }
 
 // Value parses the text content of the entry as a float64.
-// It returns the parsed float and an error if parsing fails.e
+// It returns the parsed float and an error if parsing fails.
 func (e *NumericalEntry) Value() (float64, error) {
 	t, err := e.makeParsable(e.Text)
 	if err != nil {


andydotxyz:
Can you please outline why this should be a new widget? Adding localisation is a core feature of many widgets and the core toolkit - so it doesn't make sense to me that there would be an old widget that exists explicitly for the purpose of not being localised.

jimorc:
While the changes do not break the API, there is much going on behind the scenes that is different. For example:
* NumericalEntry checks keyboard input as follows;
  * A minus sign (actually the Unicode hyphen-minus) is allowed only as the first character.
  * Any number of digits, commas and full stops are allowed.
* Any characters at all can be added/displayed when adding text via SetText, Append, and clipboard paste.
* While both commas and full stops are allowed as both thousands separators and radix separators, the user is responsible for modifying the text so that it can be parsed via `strconv.ParseFloat`.

For the new widget, in addition to localizing the input, character input is somewhat more restrictive, SetText, Append and clipboard paste all throw away invalid characters, validation of the text is added, and a ParseFloat method is added to save the user from having to modify the text before calling `strconv.ParseFloat`.

So I developed this as a separate widget to keep the functionality of the old NumericalEntry widget while greatly enhancing the functionality. I was aware that this question might come up, but it is slightly easier to edit LocalizedNumericalEntry to replace NumericalEntry than it is to go the other way in case you preferred that.

I have no problem with turning LocalizedNumericalEntry into NumericalEntry; in fact, I *half expected* that request. I am just waiting for the request to do so.

jimorc:
I renamed LocalizedNumericalEntry to NumericalEntry.

Waiting for review.

widget/numerical_entry.go
```diff
@@ -15,37 +24,131 @@ type NumericalEntry struct {
...
+// The current value will be displayed and any changes in the data will cause the widget
+// to update.
+func (e *NumericalEntry) Bind(data binding.Float) {
+	e.binder.SetCallback(e.updateFromData)
+	e.binder.Bind(data)
+}
+
+// ParseFloat parses the text content of the entry as a float64.
+// It returns the parsed float and an error if parsing fails.
+func (e *NumericalEntry) ParseFloat() (float64, error) {
```
andydotxyz, line 90: I would expect from the name "ParseFloat" that it takes a parameter to parse. Maybe "Float()" sets the expectation better? it is essentially just a getter.

andydotxyz: Or just "Value()" to reflect the "SetValue"?

widget/numerical_entry.go
```diff
@@ -72,12 +177,253 @@ func (e *NumericalEntry) Keyboard() mobile.KeyboardType {
...
-		return err == nil
+// Unbind disconnects the entry from the bound data.
+// This will remove the data updates and allow the entry to be used independently.
+func (e *NumericalEntry) Unbind() {
+	e.binder.Unbind()
+}
+
+// ValidateText checks if the entered text is a valid numerical input
+// according to the system locale.
+func (e *NumericalEntry) ValidateText(text string) error {
```
andydotxyz, line 188: It doesn't feel like this should be public API.

widget/numerical_entry.go
```diff
@@ -72,12 +177,253 @@ func (e *NumericalEntry) Keyboard() mobile.KeyboardType {
...
+
+// getRuneForLocale checks if a rune is valid for the entry,
+// and returns the correct rune for the locale.
+func (e *NumericalEntry) getRuneForLocale(r rune) (rune, bool) {
+	if unicode.IsDigit(r) {
+		return r, true
+	}
+
+	switch r {
+	case '-': // hyphen - minus
```
andydotxyz, line 243: I don't think we need comments to explain a visible rune.

p.s. all the cases that are a fall through could be on one case, shortening this a lot.

widget/numerical_entry.go
```diff
@@ -72,12 +177,253 @@ func (e *NumericalEntry) Keyboard() mobile.KeyboardType {
...
+		e.radixSep = findSeparator(runes[6:])
+		if e.radixSep == 0 && e.thouSep != 0 {
+			// If no radix separator is found, and thousand separator is present,
+			// use the last non-digit character as radix separator
+			e.radixSep = findSeparator(runes[len(runes)-7:])
+		}
+	}
+}
+
+// min returns the smaller of two integers
```
andydotxyz, line 366: Please put utility functions at the end of the file

widget/numerical_entry.go
```diff
@@ -72,12 +177,253 @@ func (e *NumericalEntry) Keyboard() mobile.KeyboardType {
...
+	t = strings.Replace(t, string(e.minus), "-", -1)
+	t = strings.Replace(t, string(e.radixSep), ".", -1)
+	return t, nil
+}
+
+// getLocaleRunes determines the minus sign, radix, and thousand separator
+// characters for a given locale by formatting a number and extracting
+// the relevant characters. It returns the minus sign, radix, and thousand
+// separator runes.
+func (e *NumericalEntry) getLocaleRunes(locale string) {
```
andydotxyz, line 322: This isn't a getter so needs a better name - it seems like like it is setup

andydotxyz (changes requested):
Thanks for this, lots of great stuff. A few comments inline.

One overall thing is that I think it won't work with &NumericalEntry{} any more because you do required setup in the constructor function.

It feels like it carries a lot of state that should be part of a lookup in the locale - the two might be related.

widget/numerical_entry.go
```diff
@@ -15,37 +24,131 @@ type NumericalEntry struct {
...
+	binder basicBinder
+	locale string
+	mPr    *message.Printer
 }
 
 // NewNumericalEntry returns an extended entry that only allows numerical input.
 func NewNumericalEntry() *NumericalEntry {
 	entry := &NumericalEntry{}
+	var err error
+	entry.locale, err = jibber_jabber.DetectIETF()
```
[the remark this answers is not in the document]
jimorc: jibber_jabber.DetectIETF() call changed to lang.SystemLocale().String() call.
Setup has been moved out of NewNumericalEntry and NewNumericalEntryWithData so that they don't do initialization that is not strictly necessary at the time they are called.

widget/numerical_entry.go
```diff
@@ -15,37 +24,131 @@ type NumericalEntry struct {
...
+
+	entry.getLocaleRunes(entry.locale)
 	entry.ExtendBaseWidget(entry)
+	entry.Validator = entry.ValidateText
 	return entry
 }
 
+// NewNumericalEntryWithData creates a numerical entry that is bound to a
+// data source and can allow or disallow float and negative numbers.
+func NewNumericalEntryWithData(allowFloat bool, allowNegative bool, data binding.Float) *NumericalEntry {
```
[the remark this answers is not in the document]
jimorc: I removed the setting of AllowFloat and AllowNegative from NewNumericalEntryWithData. If user code changes either of these fields, display is updated on next Refresh call.
I also moved the setting of onChanged into the Bind method, and set onChanged to nil in Unbind.

widget/numerical_entry.go
jimorc: Changed to Value().

widget/numerical_entry.go
jimorc: Changed to validateText.
Validator set to validateText in appropriate places if Validator is nil.

widget/numerical_entry.go
jimorc: cases combined without fallthrough comment and character comments removed.

widget/numerical_entry.go
jimorc: Moved to end of file.

widget/numerical_entry.go
jimorc: getLocaleRunes changed to setupRunes and called from setup setup was moved out of NewNumericalEntry and NewNumericalEntryWithData.

README.md
```diff
@@ -239,6 +239,141 @@ action := NewTwoStateToolBar(theme.MediaPlayIcon(),
 
 * [Demo App](cmd/twostatetoolbaraction_demo/main.go)
 
+### NumericalEntry
```
andydotxyz, line 242: This is over 130 lines of text for a single widget - in the README that covers the whole repo.
Can it be simplified - or if required to be this in-depth some of it expand into another file?

cmd/numericalentry_demo/main.go
```diff
@@ -0,0 +1,45 @@
...
+	w := a.NewWindow("NumericalEntry Demo")
+
+	nE := &xwidget.NumericalEntry{}
+	nE.ExtendBaseWidget(nE)
+	nE.AllowFloat = false
+	nE.AllowNegative = false
+	a.Clipboard().SetContent("92.f65")
+
+	valL := widget.NewLabel("Value as float:")
+	// Should call nE.SetText
```
andydotxyz, line 23: What is this comment referring to?

cmd/numericalentry_demo/main.go
```diff
@@ -0,0 +1,45 @@
...
+	nE.OnChanged = func(s string) {
+		f, _ := nE.Value()
+		valL.SetText(fmt.Sprintf("Value as float: %f", f))
+		valL.Refresh()
```
andydotxyz, line 27: Refresh is not needed after a setter method

widget/binder_helper.go
```diff
@@ -0,0 +1,79 @@
+package widget
+
+// This file is a direct copy of widget/binder_helper.go in fyne version 2.5.3.
```
andydotxyz, line 3: Why is this file required? A massive copy like this feels bad - particularly as binding code has iterated in the main Fyne repo since this.

You are extending the Entry widget so all of the binding primitive should be in already and can be used in the extended NumericalEntry

andydotxyz (changes requested):
Sorry for the long delay. I just have some suggestions inline about how this can be tidied up and I think the need for the copy of binding code in Fyne can be removed.

README.md
jimorc: Most of the description of NumericalEntry has been moved to a separate README_numericalentry.md file.

widget/binder_helper.go
jimorc: Binding in the Entry widget is to a string. The value displayed by a NumericalEntry widget is locale-specific, so all code using the bound value would have to provide similar code to convert the locale-specific string to a float. I save the user code from having to do that by binding the NumericalEntry value to a float64.

The problem is `Entry.binder` is not exported, so it is necessary to create a different binder in NumericalEntry. Because `fyne.widget.basicBinder` is also not exported, I copied the `file binder_helper.go` as the simplest solution.

Additional non-fyne widgets (i.e. fyne-x widgets and application-specific widgets) may want to provide binding in future. I will create a separate issue to address that.

andydotxyz: I think it's all getting too theoretical. Allow me to illustrate - an entry that accepts bool bindings:

```go
type boolEntry struct {
	widget.Entry
	
	bound binding.DataItem
	boundListener binding.DataListener
}

func (b *boolEntry) Bind(data binding.Bool) {
	b.bound = data
	b.boundListener = binding.NewDataListener(func() {
		loaded, err := data.Get()
		if err == nil && loaded == true {
			b.SetText("true")
			return
		}

		b.SetText("false")
	})
	data.AddListener(b.boundListener)
	b.OnChanged = func(s string) {
		parsed, err := strconv.ParseBool(s)
		if err == nil && parsed == true {
			_ = data.Set(true)
			return
		}

		_ = data.Set(false)
	}
}

func (b *boolEntry) Unbind() {
	b.OnChanged = nil
	b.bound.RemoveListener(b.boundListener)
	b.boundListener = nil
}
```

andydotxyz:
Thanks so much, my apologies for the delays

PR #119, opened 2026-03-31
Add support for map markers

sdassow:
Initial map marker implementation for Apptrix.

https://github.com/user-attachments/assets/f48509df-1f85-4365-b116-8ffe2b7734a3


sdassow:
Awesome, thanks!

widget/map.go
```diff
@@ -161,25 +190,52 @@ func (m *Map) PanToLatLon(lat, lon float64) {
...
+	return lat, lon
+}
+
+// SetMarkers updates the list of markers to show on the map.
+func (m *Map) SetMarkers(markers []MapMarker) {
+	objs := make([]fyne.CanvasObject, len(markers))
+	for n, marker := range markers {
+		objs[n] = newMapMarker(marker)
+	}
+	m.markers.Objects = objs
```
andydotxyz, line 236: This must be missing a Refresh? Setters should always ensure the new state is drawn

widget/mapmarker.go
```diff
@@ -0,0 +1,107 @@
...
+	Title() string
+}
+
+type genericMapMarker struct {
+	lat   float64
+	lon   float64
+	title string
+}
+
+func NewMapMarker(lat, lon float64, title string) *genericMapMarker {
```
andydotxyz: Returning a private type from a public API is a bad practice - won't just `MapMarker` work?

andydotxyz (changes requested):
Looking great thanks. Just 2 items inline

widget/map.go
sdassow: Good catch, thanks!

widget/mapmarker.go
sdassow: Yes, thanks.

andydotxyz (approved):
Fantastic addition thanks!

PR #120, opened 2026-05-29
widget: make CompletionEntry.FocusGained overridable

laszukdawid:
Lets a user subclass `CompletionEntry` and override `FocusGained` (the standard fyne way to react to focus) without breaking the completion popup.

The popup previously resolved its canvas and position against the embedded `CompletionEntry` value, which is not the object registered in the canvas when the widget is extended — so the lookup failed/mispositioned for subclasses. It now resolves against the extending widget recorded via `ExtendBaseWidget`, mirroring fyne's own `Select`/`SelectEntry` (which use `super()` for the same reason).

Adds `TestCompletionEntry_SubclassFocusGained` covering an embedded entry whose popup must appear below the entry rather than at the canvas origin.

Supersedes the earlier `OnFocusGained` callback approach in favour of idiomatic fyne subclassing.

### DIFF
--- a/widget/completionentry.go
+++ b/widget/completionentry.go
@@ -17,12 +17,28 @@ type CompletionEntry struct {
 
 	CustomCreate func() fyne.CanvasObject
 	CustomUpdate func(id widget.ListItemID, object fyne.CanvasObject)
+
+	impl fyne.Widget
 }
 
 // NewCompletionEntry creates a new CompletionEntry which creates a popup menu that responds to keystrokes to navigate through the items without losing the editing ability of the text input.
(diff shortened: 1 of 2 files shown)


andydotxyz:
Instead of adding new APIs should we not fix the underlying issue that the FocusGained cannot be overridden as expected?

laszukdawid:
You're right, @andydotxyz, to fix a problem instead of creating a workaround! Sorry, my bad! :(

What do you think about the current proposal?

andydotxyz:
Yes, until we find a better way then copying the pattern from upstream makes sense. 
I need to test it before approval though.

andydotxyz (approved):

PR #122, opened 2026-07-03
feat(widget): add a round gauge

roffe:
A widget that displays a value within a range as a needle on a circular dial, with graduation marks, graduation labels and a numeric readout.


widget/gauge.go
```diff
@@ -0,0 +1,301 @@
...
+	// Min and Max define the range of the gauge. They default to 0 and 100.
+	Min, Max float64
+	// Value is the current value indicated by the needle, use SetValue to change it.
+	// The needle is clamped to the range but the readout shows the value as-is.
+	Value float64
+	// Title is drawn below the centre of the dial.
+	Title string
+	// Steps is the number of tick divisions around the dial, default 10.
+	// Every second tick is drawn larger and labelled with its value.
+	Steps int
```
andydotxyz: Just a niggle here - "Steps" in other APIs (like slider) mean incremental "jumps", but this is not the meaning here - perhaps there is another name that does not overlap (Marks?)

andydotxyz (changes requested):
This is very nice, thanks. Just one note about naming which I think should be adjusted before landing.

widget/gauge.go
roffe: Renamed **Steps** to **Graduations** — that's the standard instrumentation term for the marks on a gauge scale.

PR #123, opened 2026-07-11
Adding a traced button - animating an outline

andydotxyz:
Animate a trace around abutton with a shader in a glow.
It updates to reflect the button radius and primary color of the current theme.

image


andydotxyz:
(Ignore the changes to CI - that was landed on the wrong branch)

